PR #1, opened 2017-08-28
build: fix warnings

adriweb:
Those two warnings showed up while building the whole thing, it wasn't very beautiful in the terminal :)

* `escher/src/text_area.cpp:113:25`: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
* `liba/src/external/openbsd/k_rem_pio2f.c:66:54`: warning: this 'for' clause does not guard this statement, but the latter is misleadingly indented as if it is guarded by the 'for' [-Wmisleading-indentation]


adriweb:
Ah, right, I had not noticed the existing Makefile CFLAGS overrides.
Commit updated.

PR #12, opened 2017-08-29
Fix #9 by deactivating empty keys.

### LINKED ISSUE #9: Pressing empty keys causes crashs
Some keys are empty, I guess that's because the layout doesn't match the physical one and then are useless.

If any one of the empty keys are pressed, the app crash with error:
```
app.elf: escher/src/run_loop.cpp:34: bool RunLoop::step(): Assertion `event.isDefined()' failed.
```

jacobly0:


adriweb:
Works fine for me.

Issue #6, opened 2017-08-29
Rounding error with atan(acos(asin(ans)))


wiiikiii:
https://www.numworks.com/eu/simulator/

tan(cos(sin(9)))
atan(acos(asin(ANS)))

the result is 8.989918

is this result only in simlulation or on hardware also?

rhaamo:
I tried the web sim and with the built one and I can only get "undef"

How to reproduce that ?

![selection_038](https://user-images.githubusercontent.com/8625/29827630-bf9a8214-8cda-11e7-8480-4cc9947b6dd2.png)

OLaurent-Maths:
Wrong order of the functions in the issue description:
asin(acos(atan(ans)))

rhaamo:
Thanks! I'm getting the same with the built simulator.
I will try tomorrow with the real calc if there is the same issue.

rhaamo:
I've done more "tests", sim vs "another calc"
```
sin(9), 0.156434 vs  0.412118
cos(9), 0.987688 vs -0.911130
tan(9), 0.158384 vs -0.452315
```

Using KCalc on KDE, I got the same results as the simulator (web or "desktop app"), maybe this could be related to endianess of the platform.

OLaurent-Maths:
Works fine for me on another calculator (TI)
My guess is that when calling ANS, it tried to parse again the previous result : 0.017455 instead of getting the complex number behind it.
If you just try to do: asin(acos(atan(tan(cos(sin(9))), it works perfectly

rhaamo:
Still imprecise.
A casio responds with "0.4247779608", the sim "9", a single cos(), tan() or sin() is even wrong in the sim.
About the wront re-usage of "Ans" I think this is a separate issue.

OLaurent-Maths:
"A casio responds with "0.4247779608", the sim "9",-> this is normal : your CASIO calculator is set to radian so 0.4247779608 is also correct.

Ecco:
Actual results from a device running `679CEA7`.

`asin(acos(atan(tan(cos(sin(9)))))` gives `9`

and

```
tan(cos(sin(9)))
-> 0.017455
asin(acos(atan(ans)))
-> 8.989918
```

This is indeed a rounding bug.

OLaurent-Maths:
Indeed, when evaluating the 'ans' value (Evaluation<double> * LocalContext::ansValue()) it takes the last result of the calculation store with the precision of 7 digits.

[...]

PR #15, opened 2017-08-29
Fixed precision when using ANS: Issue #6

OLaurent-Maths:


EmilieNumworks:
The solution would be to store the input expression 'tan(cos(sin(9)))' in 'ans' instead of the output '0.017455'. However, in that case, if the user changes the mode (from degree to radian for instance), the 'ans' would not be equal to '0.017455' as expected but to '1.303162' (the result of 'tan(cos(sin(9)))' in radian)...

OLaurent-Maths:
Thanks for you comments.
It's true we cannot store directly the content to ans, especially if there are cascaded calculations.

I propose in the output to recompute directly from the inputText.
It will fix the issue of going back to the Main Menu, and also it will fix the problem if the user switches from degrees to radians.
I'll send a patch right now

jacobly0:
Not having looked into the code yet, the fact that it works if you store the intermediate value into a variable like X suggests that Ans should be implemented in a similar manner.  I supposed there may be a typing issue because A-Z appear to only hold numbers (a feature I'm not a huge fan of to begin with), not e.g. matrices, but Ans certainly needs to be able to hold any type.

OLaurent-Maths:
See my previous comment, but for instance if you do this : 
tan(9)
cos(ans)
sin(ans)

You cannot store a single value/object into ans, since it relies on the previous operations on the stack

jacobly0:
I was under the impression that @EmilieNumworks was worried that if changing modes changed ans then users would easily become confused by the fact that ans is now a different number that what is displayed on the previous line.  It seems like the point would be to make ans *not* dependent on previous operations, and simply be whatever information would have been stored in the variable X had that line ended with say ...&rightarrow;X, so that logically each line ends with ...&rightarrow;ans where ans is a variable that can hold any type.

OLaurent-Maths:
Ok I see your point. So to summarize:
The m_outputObject is deleted for heap management when switching app: so one way or another, when switching back to calculation app, it has to be recomputed:
- it can be recomputed from the m_outputText (what is currently done), however this means a precision loss (7 digits only)
- it can be recomputed from the m_inputText, however this means that the whole calculation is done again: it allows to make a good use of ANS, however it will change the expected result if the user switched from degrees to radians.

One idea could be:
At Calculation creation, store locally the angle unit within the calculation object
At output, instead of evaluating expression like this : exp->evaluate<double>(*context); , do it by passing the angle unit specific to this calculation : exp->evaluate<double>(*context, angleunit);

jacobly0:
An interesting point that occurs to me is that if you were to compare using ans and pasting the previous line, the latter is going to give you the approximation anyway.  Therefore, the only information we really need to keep around about older (all except the previous) calculations is the input/output strings themselves, which as I understand is what is currently done.
However, I can't think of any more reasonable way to remember the accurate ans value than by keeping the previous (and only the previous) evaluation around, even after changing apps.  If you were to execute `tan(9)`, change mode, execute `cos(ans)`, change mode, execute `sin(ans)`, etc. you would end up needing to remember all the settings for each calculation, which seems it would waste more memory than just the one evaluation, which we need to create anyway.
As for where to store it, I imagine you can either avoid deleting the most recent m_output on app changes (you can probably delete the previous ones as soon as they are no longer the most recent), or store it somewhere in the context chain.

OLaurent-Maths:
So when switching app, instead of cleaning up all the calculations , we could only clean up all but the last calculation. 
So the last result would be stored in the Calculation.m_output itself, no need to find another place to store it.

apps/calculation/calculation.cpp
```diff
@@ -52,9 +52,16 @@ void Calculation::reset() {
 void Calculation::setContent(const char * c, Context * context) {
   reset();
   strlcpy(m_inputText, c, sizeof(m_inputText));
-  Evaluation<double> * evaluation = input()->evaluate<double>(*context);
-  evaluation->writeTextInBuffer(m_outputText, sizeof(m_outputText));
-  delete evaluation;
+  if(m_output==nullptr) {
```
EmilieNumworks: reset() calls tidy() which ensures that m_output == nullptr
So this patch could be only:
assert(m_output == nullptr);
m_output = input()->evaluate\<double\>(*context);
m_output->writeTextInBuffer(m_outputText, sizeof(m_outputText));

apps/calculation/calculation_store.cpp
```diff
@@ -84,8 +84,13 @@ void CalculationStore::deleteAll() {
 }
 
 void CalculationStore::tidy() {
```
EmilieNumworks: m_calculations is a ring buffer, there is no guarantee that m_calculations[numberOfCalculations()] is the last added one.

EmilieNumworks:
With a bit of hindsight, this patch would not fix the issue if we introduce ans[i] later that gives the last i th answer. We can maybe extend the number of significative digits that are stored (and not printed) in calculation. You can check my branch 'precision-ans', I did not have time to check it but the idea is there.

OLaurent-Maths:
I tested your branch and it seems to work.
By separating printed/stored digits, we could also add an option in the Parameters to configure the number of printed digits.

EmilieNumworks:
Mmm it works but only for the simulator so far. However, you cannot build that for the device as I did not implemented conversions from long to float or long to double yet (aeabi_l2f, aeabi_l2d etc). I add that to my TODO list as it would also be required for the issue #40 you raised.

OLaurent-Maths:
Also, I think, if later on we want to be able to change the number of printed significant digits:
we would need to change the inputLayout and ouputLayout methods.
For instance, in the input, we do not want to round the expression, but we do for the ouput:
if we set 2 significant digits (one decimal), and we type 0.15, it should display in the calculation:
0.15
                   0.2
, meaning that the input is not affected/rounded because of the precision settings

I guess we have to also pass a numberofSignificantDigits in the Expression::createLayout method

EmilieNumworks:
I close this pull request as it is connected to #61.

PR #21, opened 2017-08-30
Send a termination event when the simulator is closed.

jacobly0:
Having an arbitrary pending event may be too general, but in the simulator `window->callback` should be connected to generating a termination event one way or another.


ion/src/shared/events_keyboard.cpp
```diff
@@ -29,6 +29,13 @@ bool canRepeatEvent(Event e) {
 Event getEvent(int * timeout) {
```
EmilieNumworks: This function is called in the software running on the device. We try to avoid adding code that would be useful only for one platform when this code is built for all platforms. Indeed, we do not want to downgrade performance on the device to improve a functionality of the simulator.

EmilieNumworks:
The callback is a great idea but we have to update pendingEvent outside of getEvent() (which is used on all platforms - device, simulator and emscripten).

jacobly0:
By moving `Ion::Events::getEvent` and other keyboard specific code into the namespace `Ion::Keyboard`, we can free up `Ion::Events::getEvent` to be implemented differently on each platform.  The device version simply delegates to `Ion::Keyboard::getEvent`, but the simulator version also checks if a termination event should occur.  The only way I see of avoiding the extra call on the device is to check the platform with preprocessor conditionals or completely duplicate the code in `Ion::Keyboard::getEvent`.

adriweb:
That looks pretty nice :)

jacobly0:
It occurs to me that a better solution may be to have a customized input routine for the simulator à la the emscripten platform, since the current way is a bit clunky in comparison, which would make the above issue moot since the code would no longer be the same.

adriweb:
Note: I too recommend #59 over this one :)

Ecco:
Superseeded by #59. Closing 😄

PR #27, opened 2017-08-30
[ion] [simulator] more realistic button labels

adriweb:
Looks like this now:

screenshot


adriweb:
Updated with better arrows (thanks @jacobly0)

Ecco:
That's a very nice addition, thank you so much!

That being said, I'm not comfortable using non-ASCII characters in the source code. I know it shouldn't be an issue with most toolchains, but this may end up triggering weird corner-case issues down the road.

Would you mind updating your patch to use Unicode escape codes (\uXXXX)?

adriweb:
Alright, this should probably be fine (couldn't test right now though).

Ecco:
Looks great. We fixed a typo on the CLA (a missing space…), so we would need you to sign it before we can merge this…

PR #39, opened 2017-08-30
[escher] table view: ability to jump to top/bottom

adriweb:
When you're at the top of a list, you can press [up] to go to the bottom instantly, and same with [down] to go to the top when at the bottom.
Similarly, works with columns with left/right.

This feature is only enabled when it's 2x2 or more because otherwise:
 - it may cause confusion
 - it would break the specific behaviour of submenus.

(I haven't tested it much, but so far so good... Please try it and report back :D)

Screenshot:
On the math toolbox

### DIFF
--- a/escher/src/selectable_table_view.cpp
+++ b/escher/src/selectable_table_view.cpp
@@ -60,11 +60,29 @@ void SelectableTableView::deselectTable() {
 }
 
 bool SelectableTableView::selectCellAtLocation(int i, int j) {
-  if (i < 0 || i >= dataSource()->numberOfColumns()) {
-    return false;
+  int nbrColumns = dataSource()->numberOfColumns();
+  int nbrRows    = dataSource()->numberOfRows();
+  if (i < 0) {
+    if (nbrColumns <= 2) {
+      return false;
+    }
+    i = nbrColumns-1;
+  } else if (i >= nbrColumns) {
+    if (nbrColumns <= 2) {
+      return false;
+    }
+    i = 0;
   }
-  if (j < 0 || j >= dataSource()->numberOfRows()) {
-    return false;
+  if (j < 0) {
+    if (nbrRows <= 2) {
+      return false;
+    }
+    j = nbrRows-1;
+  } else if (j >= nbrRows) {
+    if (nbrRows <= 2) {
+      return false;
+    }
+    j = 0;
   }
   unhighlightSelectedCell();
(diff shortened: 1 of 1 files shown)


artaxxx:
Actually this is a feature we wanted to include in the calculator when we designed it.

Our beta testers found it a little bit confusing and we chose to remove it to keep a real "geographical" navigation.

Did you test it on the device? What do you think?

adriweb:
I don't have a physical device myself (yet, anyway), so I've only tried on the simulator.
I think it not having that feature bothers me way more considering that on lists such as the math toolbox I have to scroll all the way to the bottom, for instance.
I do agree it can be confusing when table have a short number of items, which is why my code only enables it on greater than 2x2: you can clearly see which ite, is the currently selected one, such that moving doesn't appear as "toggling".

What do other people think?

Ecco:
It looks like a great addition, but I think you'd agree it's not if you had an actual device in hand.

Thing is, the actual device has an auto-repeat feature on arrow keys which makes it **a lot** easier to browse menus. Whatsoever, having "looping" menus with the auto-repeat feature makes it actually a lot **harder** to navigate.

So unfortunately I'm going to close this pull request. We made a small video, I hope it'll have you convinced.

On the math toolbox

boricj:
How about keypad shortcuts in menus, like in the TI-83 series? Hit 1 to select/use the first item, hit 2 for the second and so on. It'd be faster for experienced users but without drawbacks for beginners.

Ecco:
Yeah, that sounds like a good suggestion. Don't hesitate to open another issue :)

adriweb:
(Indeed, the videos shows more speed/smoothness than what I initially imagine, so it's not _that_ bad having to scroll)

On IRC, we've also agreed that it'd be good to have shortcuts like [shift][arrows] to quickly jump around (but not only just in lists, on text inputs too).
This has the advantage of not modifying the default no-auto-jump behaviour. I suppose I could try to take a look at that this week-end..

@boricj yep, +1. Probably for another PR than what I propose here ^ though

blackketter:
@boricj Another suggestion: predictive typing using the alpha characters.  It would be awesome to hit TOOLBOX R O O to filter down root().  (I know this belongs in the new issue, but i didn't see it yet... :)

matheod:
Why not mix both ? If you hold down, it will quickly go to down and stop, but if you press once again, it will go top.

PR #51, opened 2017-08-31
Add .DS_Store to ignore list

urashid64:
For cleaner git status on Mac OS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ escher/image/inliner
 
 # Ignore generated qtsrdefs file
 python/port/genhdr/qstrdefs.generated.h
+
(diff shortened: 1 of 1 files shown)


Ecco:
Hi @urashid64, thank you very much for this contribution. I would like to apologize but unfortunately we're not going to merge this patch.

Indeed, `.DS_Store` are not specific to our project but rather to an OS (macOS in this case). The proper way to ignore them would be to add a global git ignore. See https://help.github.com/articles/ignoring-files/ and https://gist.github.com/octocat/9257657

urashid64:
Fair enough. Thanks for the reply.

PR #52, opened 2017-08-31
[apps/code] template program: don't memcpy from beyond the string

adriweb:
- Was caught by ASan
- Added a static_assert on the length just in case
- Big C strings concats replaced by raw C++ literal for clarity


EmilieNumworks:
Great patch! Thank you.

PR #54, opened 2017-08-31
Fix swapped matrix transpose, multiplication, opposite, and clone dimensions

jacobly0:
This adds matrix dimension checking to the tests, which reveals a bunch of similar bugs.  I searched for any instances of col before row, and changed all the ones that were either bugs or didn't have a reason to be in that order.  The api breaking changes were MatrixData and ExpressionMatrix constructors swapping the order of the dimension parameters.


EmilieNumworks:
@jacobly0, this pull request is excellent, thanks a lot!

PR #57, opened 2017-09-01
Automated testing on travis using the blackbox.

jacobly0:
Also increased test output to make it easier to determine what test failed.


EmilieNumworks:
This looks great! Very helpful for spotting the test you break.
However, for next time, could you pull smaller branch? That way it would be easier to review and to merge. For example, separating the part where you just clean poincare/test, from the part you add a console option to tests, from the part you add command to travis.

PR #59, opened 2017-09-01
Custom event handling for the simulator.

### LINKED ISSUE #8: Closing the window doesn't exit
Closing the windows will leave "app.elf" (or whatever named is) running, needed to be manually killed.

jacobly0:
This is an alternative to #21 that also fixes #8.


jacobly0:
This PR also has the side effect of reducing the idle simulator cpu usage from 95% to 1% on my computer, something which #21 does not do.  Also, if it's not obvious this alternative has zero cost for the device platform.

adriweb:
Great changes.
CPU usage completely dropped on mac here and it definitely feels snappier (responding to keypresses "better" etc.).

jacobly0:
Rebased.

Ecco:
Excellent PR. Thank you very much @jacobly0 !

PR #61, opened 2017-09-01
[apps/calculation] Increase the precision of the output text stored in every calculation

EmilieNumworks:
this enables to have more precision for symbol ANS


madhatter0:
Can I beg that this fix be pulled?  It addresses the single longest-standing bug in the tracker (#6).  I'd really, really like to get a high-precision "Ans" into the mainstream.

EmilieNumworks:
I close this, it is obsolete with 1.2.0. Sorry for the delay.

PR #65, opened 2017-09-01
[apps/settings] sub controller: fix buffer overflow.

adriweb:
- Caught by ASan
- The 3 and 4 length were swapped
- It's now using compile-time automatic sizes

### DIFF
--- a/apps/settings/sub_controller.cpp
+++ b/apps/settings/sub_controller.cpp
@@ -20,11 +20,11 @@ SubController::SubController(Responder * parentResponder) :
     m_cells[i].setAccessoryFontSize(KDText::FontSize::Small);
     m_cells[i].setAccessoryTextColor(Palette::GreyDark);
   }
-  const char text[6] = {'a','+', Ion::Charset::IComplex, 'b', ' ', 0};
+  const char text[] = {'a','+', Ion::Charset::IComplex, 'b', ' ', 0};
   m_complexFormatLayout[0] = new StringLayout(text, 6);
-  const char base[3] = {'r', Ion::Charset::Exponential, 0};
-  const char superscript[4] = {Ion::Charset::IComplex, Ion::Charset::SmallTheta, ' ', 0};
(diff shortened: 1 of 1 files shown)


Ecco:
Great find, thanks!

apps/settings/sub_controller.cpp
```diff
@@ -20,11 +20,11 @@ SubController::SubController(Responder * parentResponder) :
...
   }
-  const char text[6] = {'a','+', Ion::Charset::IComplex, 'b', ' ', 0};
+  const char text[] = {'a','+', Ion::Charset::IComplex, 'b', ' ', 0};
   m_complexFormatLayout[0] = new StringLayout(text, 6);
-  const char base[3] = {'r', Ion::Charset::Exponential, 0};
-  const char superscript[4] = {Ion::Charset::IComplex, Ion::Charset::SmallTheta, ' ', 0};
-  m_complexFormatLayout[1] = new BaselineRelativeLayout(new StringLayout(base, 4), new StringLayout(superscript, 3), BaselineRelativeLayout::Type::Superscript);
+  const char base[] = {'r', Ion::Charset::Exponential, 0};
+  const char superscript[] = {Ion::Charset::IComplex, Ion::Charset::SmallTheta, ' ', 0};
+  m_complexFormatLayout[1] = new BaselineRelativeLayout(new StringLayout(base, sizeof(base)), new StringLayout(superscript, sizeof(superscript)), BaselineRelativeLayout::Type::Superscript);
```
adriweb: @Ecco I just noticed that while this patch does indeed fix the overflow crash, this sizeof is a bit overzealous considering StringLayout already allocates len+1. So, in theory, it could also just be `sizeof()-1` to save a byte of heap. Or better: we could remove the 0 at the end of each char arrays and leave the sizeof() as is.
Considering I've seen another piece of code that has an overflow like this one, I'll fix it too with that in mind.

PR #81, opened 2017-09-04
Repair windows build

EmilieNumworks:


adriweb:
(typo: "repair", or just "fix")

@jacobly0 you might want to check this out, compared to [your windows branch](https://github.com/jacobly0/epsilon/tree/windows).

adriweb:
Shall I suggest cmake? :P

rhaamo:
CMake will clearly not be easier than Makefiles, also modularity won't be too.
Without talking about managing the build for device or emscripten or simulator or blackbox.

For detecting windows in Makefile it's possible to use something like this (https://stackoverflow.com/a/12099167):

```makefile
ifeq ($(OS),Windows_NT)
    #Windows stuff
    ...
else
    #Linux stuff whatever
    ....
endif
```

adriweb:
BTW, on Windows, building works fine in WSL according to a reddit comment (however, I don't think the simulator has been launched in it, since it would require X11 or something)

Ecco:
> Shall I suggest migrating to cmake?

You sure can, but we'd really need a compelling reason. I'm pretty sure GNU Make can do the job :)

Ecco:
> BTW, on Windows, building works fine in WSL according to a reddit comment (however, I don't think the simulator has been launched in it, since it would require X11 or something)

That's good to know. The big upside of using mingw is that you can produce a binary that runs on any Windows computer, without any dependency. Which is definitely something that's useful :)

Ecco:
> For detecting windows in Makefile it's possible to use something like this (https://stackoverflow.com/a/12099167)

Oh, that's a good find. Actually we could use it in `platform.simulator.mak` to automagically pick the mingw toolchain. Great suggestion, thanks @rhaamo :)

Ecco:
Ok, let's merge this as-is for now, since it's a good first step anyway.

PR #82, opened 2017-09-04
Removed the automatic insertion of 'ans' when pressing '-'.

mjmdavis:
My girlfriend was using the calculator and was becoming confused when it kept entering `ans-` when she pressed`-`. She was trying to do basic calculations with negative numbers and was getting held up.

Perhaps in future this could be a modal feature but for now I suggest removing the addition of `ans` when `-` is the first character pressed.

### DIFF
--- a/apps/calculation/text_field.cpp
+++ b/apps/calculation/text_field.cpp
@@ -20,7 +20,6 @@ bool TextField::handleEvent(Ion::Events::Event event) {
   if (textLength() == 0 &&
       (event == Ion::Events::Multiplication ||
        event == Ion::Events::Plus ||
-       event == Ion::Events::Minus ||
        event == Ion::Events::Power ||
        event == Ion::Events::Square ||
        event == Ion::Events::Division)) {


adriweb:
Heh, this is precisely why most calcs have two, distinct, keys for negation and subtraction...
Having to chose one over the other by default is not going to be easy :/

mjmdavis:
This PR seems like a good compromise for now. The calculator is more usable and less confusing.
I do think that it is a little inconsistent though so there are some other options for a long term solution.

* Merge this PR and drop the matter.
* Make the device modal and have a *running calculation* mode.
* Have a hidden negation button 😒.
* Implement collections (arrays)

I wonder if it might be better to encourage users to avoid *running calculations* in general and push them towards using collections or the statistics or regression modes. I don't know if collections would cover all the use cases of *running calculation* but it would be a nice way to introduce students to vectorized operations and maybe linear algebra. It would also allow for reproducibility and repeatability while enabling the user to check and edit their entry.

jacobly0:
One interesting point is that if you are doing a running calculation, you aren't going to want type negative except at the beginning or after an operator, where it's not ambiguous.  Similarly if you aren't doing a running calculation you don't generally need the automatic ans insertion anyway.
My suggestion is to have a separate app called, say, "Calculator" which is more reminiscent of a classic 4-function calculator.  Typing `4` `+` `5` `+` would update a static location with `9` until you start typing the next number.  Unary functions like &radic; would replace the currently displayed value with its square root, and clear would clear the calculation allowing you to start the new calculation with a potentially negative number.  I imagine you could use the rest of the screen to display the whole current calculation, and maybe have separate lines for current input and current result.  I notice that the current windows calculator behaves similarly to this and also displays a history of recently executed calculations.
This would remove the need to have the Calculation app auto insert ans, where you can still just type it manually when you actually need it, and also improves the usability of the calculator for people used to non-graphing calculators.

mjmdavis:
As a sanity check, my girlfriend, who is learning maths, thought that it seemed reasonable to not insert `ans` before `-` but to do it before all the other expressions.

Ecco:
@mjmdavis And she has a good point: "*5" or "/4" don't make any sense. "+3" barely does. Whereas "-6" definitely means something.

mjmdavis:
I just pulled my calculator out. It seems that what is missing for me is an `AC` or `All Clear` button. I have internalized pushing that button when I want to start a new calculation.

Another option would be to modify the behavior of the `clear` key to also clear or place a break in the calculation history.

adriweb:
Edge case: [-0](https://en.wikipedia.org/wiki/Signed_zero) (but let's just ignore that...)

Ecco:
This is a great discussion, thank you for starting it ! There are indeed quite a few solutions to this issue, but we may as well just try @mjmdavis 's code and see how that goes.

Thanks everyone!

PR #85, opened 2017-09-04
[various apps] StringLayout: fix buffer overflow and use auto size

adriweb:
Following a comment of mine [here](https://github.com/numworks/epsilon/pull/65/files#r136870515)...

* Another buffer overflow was happening when changing the complex format in the settings (instant crash with ASan)
* All const char arrays manually ending with a nul char were trimmed of it considering StringLayout allocs one already.
* Use automatic compile-time array-size when possible, it avoids mistakes.

I was tempted to make a macro like `#define AutoStringLayout(charArray) new StringLayout(charArray, sizeof(charArray))` but oh well :)


EmilieNumworks:
Very good point!

PR #86, opened 2017-09-05
[poincare] Implement square root of negative real

ian-abbott:
The current implementation of square root calls 'std::sqrt' if the
parameter is real, otherwise it uses calls Poincare's 'Power::compute'
with an exponent of 0.5.  Change it to only call 'std::sqrt' if the
parameter is a non-negative real number.


adriweb:
Looks like this got fix a few minutes ago actually ;) 
https://github.com/numworks/epsilon/commit/9f7b01954cae

ian-abbott:
Closing as there is already a fix in the 'root' branch.

Ecco:
Thanks for your help @ian-abbott . I'm sorry we landed the patch just a few minutes before your submission.

Issue #87, opened 2017-09-05
[Poincare] Odd integer powers of negative real numbers give positive result


ian-abbott:
For example: `(-2)^3` results in 8 instead of -8.

ian-abbott:
Also, negative integer powers of negative real numbers give a positive result. For example (-2)^(-2) results in 0.25 instead of -0.25.

Ecco:
Fixed by #90 . Thank you @ian-abbott !

PR #89, opened 2017-09-05
Fix  issue #87: Integer powers of negative real numbers

ian-abbott:
A negative real number raised to a negative integer power is negative.
A negative real number raised to an odd integer power is negative.


EmilieNumworks:
There is indeed something to do here to fix (-2)^3! However, we did not implement std::fmod in the small libc imported on the device. 
Furthermore, the cases handled separately for 'power' may be useless now (they were useful when we were computing in float instead of double)... So you just might get rid of them.

ian-abbott:
Build failure due to unimplemented std::fmod. Will close and rethink!

ian-abbott:
Half of my 'fix' was incorrect anyway! A negative real number raised to a negative integer power isn't always negative!

PR #91, opened 2017-09-05
[python] Fix type of n_args parameter of mp_builtin_open()

ian-abbott:
The n_args parameter of mp_builtin_open() has type size_t in
"python/src/py/builtin.h", and type uint in "python/port/port.c".
Change it to size_t.


Ecco:
Good catch. Thank you @ian-abbott !

PR #101, opened 2017-09-07
Fix #100.

### LINKED ISSUE #100: 0^2 returns undef
In 1.1.1 0^2 returns undef in Calculations, Functions, and Sequences (I did not test in the other apps)
It worked as expected under 1.1.0, so it must be a recent change that caused the undesirable behavior.

It works OK in python.

jacobly0:


EmilieNumworks:
Thank you very much for your commit (great adding a test)! I am merging it right away to fix the bug.

PR #106, opened 2017-09-07
[liba] Add a bridge for windows build

EmilieNumworks:


liba/include/bridge/alloca.h
```diff
@@ -0,0 +1 @@
+#include <malloc.h>
```
jacobly0: This should be guarded on windows (`#ifdef _WIN32` or something) and you should fall back to `#include_next <alloca.h>` otherwise.

Ecco: Indeed, excellent suggestion @jacobly0 . Thank you!

Ecco:
Closing this as it is part of #239 .

PR #107, opened 2017-09-07
Makefile: Use POSIX shell redirection syntax for app_flash target

ian-abbott:
The recipe for the 'app_flash' target uses an 'until' loop to wait for
the DFU device to be connected.  The command uses &> to redirect both
standard output and standard error output to a file (/dev/null).  This
works for shells such as Bash and Z Shell, but not dash or Korn shell.
Replace it with a combination of > and 2>&1 for portability.


Ecco:
Indeed, good suggestion. Thanks @ian-abbott

PR #112, opened 2017-09-09
Fix arg(0) == undef

jacobly0:
While arg(0) can return any angle and still be valid, the usual convention is to define it as 0 which makes computations easier.


EmilieNumworks:
The convention arg(0) = 0 does not seem universal (do you have any reference?) so I rather let it undefined as it is an indeterminate form.

jacobly0:
I must have been thinking of programming conventions.  You are correct, it is indeed undefined mathematically.

PR #113, opened 2017-09-09
Fix #102 Power returns undef when it has multiple possible values

### LINKED ISSUE #102: Power returns undef when it has multiple possible values.
i^i computes as undef, presumably because it can have multiple values all of the form e^(2k&pi;-&pi;/2) for any integer k.  However, every calculator I've tested just uses e^(-&pi;/2), the value on the principal branch, which seems perfectly reasonable to me considering that &radic;(4) also has multiple possible values, -2 and 2, even though everyone expects it to return the principal value of 2.  I'd be happy to try to fix this, but I want to make sure I'm understanding the issue correctly first.

jacobly0:
Fixes #102, depends on #112 to not break preexisting tests.


EmilieNumworks:
This enhancement looks great, thank you! However, as we rather let arg(0) = undef, there are few special cases to handle.

jacobly0:
I added a special case for a 0 base.

EmilieNumworks:
Thank you very much!

PR #114, opened 2017-09-09
Fix potential overflow in complex radius calculation.

jacobly0:
This can be observed by computing abs(1E155i) which returns inf even though the actual result of 1E155 easily fits in a double.


EmilieNumworks:
Once again, good mathematical trick! Thanks.

PR #116, opened 2017-09-09
shift alpha lock view: fix gcc warning about a missing return in numberOfSubviews

adriweb:
ARM GCC 6.3 was complaining about it, for some reason.


mateoconlechuga:
I purposely left this as it requires the default state. You can add another case; but leave the `Ion::Events::ShiftAlphaStatus::Default` state alone.

adriweb:
Heh, it is exactly the same, but if it makes you happier... here's an updated version so that gcc doesn't whine about the missing return (which shouldn't actually be needed... who knows why it doesn't see it)

mateoconlechuga:
Woop :+1:

PR #118, opened 2017-09-11
[apps/regression] Fix typo

tjhorner:
Fixes a simple typo in the Regression app when there's not enough data:

"Not enough data for reg**er**ssion" -> "Not enough data for reg**re**ssion"


Ecco:
Very good find @tjhorner! The PR itself is a bit weird though: would you mind removing the Merge commit?

tjhorner:
If I did that, then my fork would not be equal with the base fork, so it wouldn't be mergeable (it would have old code instead). I initially forked it a while ago, so that's why it's there. (Unless I'm just bad at git hah)

boricj:
You can use `git rebase` to rewrite history and, among other things, remove unnecessary merge commits. Rebasing on top of NumWorks's master branch will get rid of it.

Beware, rewriting history on already published remote branches requires force pushing, but that's not an issue for ephemeral branches typically used in pull requests.

adriweb:
@Ecco you could simply do a squash merge, I believe.

Ecco:
Ok, let's give the squash merge a go :)

Ecco:
Thank you @tjhorner

PR #119, opened 2017-09-11
Fix potential overflow in complex division calculation.

jacobly0:
This can be observed by computing 1E155/1E155(1+i) which returns inf even though this is equivalent to 1/(1+i) == (1-i)/2 == .5-.5i

### DIFF
--- a/poincare/src/fraction.cpp
+++ b/poincare/src/fraction.cpp
@@ -3,7 +3,7 @@ extern "C" {
 #include <string.h>
 #include <float.h>
 }
-
+#include <cmath>
 #include <poincare/fraction.h>
 #include <poincare/multiplication.h>
 #include "layout/fraction_layout.h"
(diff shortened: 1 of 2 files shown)


EmilieNumworks:
The mathematical trick is great!! However, I would rather let the special case for pure real numbers as there are still more operations (likely to propagate error on float) in your version than in only doing c.a()/d.a(). Also, c/0 with c a real would result in 'undef' instead of 'inf'.

jacobly0:
A surprising benefit to using this method is that it doesn't propagate *any* error on exactly real inputs.  Everywhere else, real inputs are checked with `d.b() == 0`, so we can assume that b will be precisely 0 (not close to 0) for real inputs. Now `abs(d.b) <= abs(d.a) == 0 <= abs(d.a)` is always true so then `temp = min/max == d.b/d.a == 0/d.a == 0` precisely because floating point 0/x is defined to return exactly 0 for normal x.  Now `norm = temp*min + max == temp*d.b + d.a = 0*0 + d.a == d.a` precisely because floating point `0*x` is defined to return exactly 0 for normal x and `0+x` is defined to return exactly x for any x.  Now `result = (temp*aa + ab) / norm, (temp*bb + ba) / norm == (0*c.b + c.a) / d.a, (0*-c.a + c.b) / d.a == c.a/d.a, c.b/d.a` precisely for reasons already mentioned.

I agree that the above breaks down only when `d.a == d.b == 0` but I'm not sure how comfortable I am with (-)r/0; r&ge;0 evaluating to (-)inf on a calculator without a limit being involved in the calculation, especially when we aren't differentiating between 0 and -0 in results (whereas with a limit, you would have to specify direction explicitly anyway).  Certainly before I learned about limits in school, I was taught that r/0 was undefined.  Also note that lim x&rightarrow;0 r/x is, in fact, undefined, but lim x&rightarrow;0<sup>+</sup> r/x; r&ge;0 == +inf and lim x&rightarrow;0<sup>-</sup> r/x; r&ge;0 == -inf.

EmilieNumworks:
Oh nice! No error propagation on real numbers then!
And you are right, 1/0 = inf is almost wrong in a way, especially for high school students. I am merging that, thank you.

Issue #42, opened 2017-08-31
Will we be able to add applications coded in C/asm?


Zezombye:
On Casio calculators, you can code an application (called addin) in C (or any language that compiles to asm), that is transferred on the storage memory and disabled in exam mode. I think it's the same on TI.

This feature would be very useful, because while some applications would benefit being incorporated into the OS (such as Eigenmath, geometry app, physics app), since the storage memory isn't unlimited it'd be better if the users could choose which apps they could add. And I don't think you'd accept my pull request of my sokoban game coded sloppily :p

Also, you could add a slot for an SD card in order to add more apps, if the flash is small.

Ecco:
This is actually a distribution issue since you can already add C/asm apps to your own device: just add one in Epsilon's sources. Now obviously we're not going to merge every single app in Epsilon, so we definitely need a way to distribute 3rd-party apps.

mateoconlechuga:
In addition to this; because of the limited flash space I think it may be beneficial to allow a more modular approach to applications. Rather than having them embedded in firmware; it would be neat to be able to add/delete them on the fly. Anywho, just an idea.

boricj:
I don't think adding native apps at run-time would be a good idea since we can just freely recompile the firmware from source and do whatever we want. What we need is a modular way of managing apps at compile-time (going as far as being able to unselect official apps for additional Flash space), perhaps even some sort of third-party hub so as not to hunt down community apps scattered across the Internet. Let people build their own calculator, in some sort.

If people really want the ability to add apps at run-time I'd go for packaged apps in MicroPython bytecode, but even then I'm not sure if this would be a good idea.

rhaamo:
Adding apps at runtine would be hard without SD card support or a builtin secondary non-volatile flash storage.

About modularity on compile-time, what's actually in place isn't a lot, there is hardcoded number of apps, manually adding hooks etc.

For what @boricj is saying about a "hub", that would means some refactoring to just have to do something like "include myapp/Makefile" in "apps/Makefile" and it would be "magically" added.

Zezombye:
Adding apps at runtime should be easy actually, Casio does it fine (for TI I don't know). When you go to the menu (or when you add something to storage memory) the system scans for application files (which could be .app or .nwk for example) and automatically add them to the menu.

Having to recompile the whole OS would be harder, I currently have spent 5 hours trying to build it, still not done yet. It would be a lot easier if you could just transfer a .nwk file to the calculator, and it appears automatically in the menu.

[...]

PR #131, opened 2017-09-14
Selection of apps at compile-time

Ecco:
This pull request allows to pick (and order) apps at compile-time.

It is a necessary step for #42 and a reimplementation for #73.

### DIFF
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -14,6 +14,7 @@ include apps/code/Makefile
 
 app_objs += $(addprefix apps/,\
   apps_container.o\
+  apps_container_storage.o\
   apps_window.o\
   battery_timer.o\
   battery_view.o\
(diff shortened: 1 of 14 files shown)


apps/Makefile
```diff
@@ -1,14 +1,20 @@
...
-include apps/probability/Makefile
-include apps/regression/Makefile
+include apps/hardware_test/Makefile
+snapshots =
+
+# You can select below which apps you want to include in your build, as well as
+# their order on the home screen. Each Makefile below is responsible for setting
+# the $snapshots variable (name of the snapshot class) and the $snapshot_headers
+# (path to the snapshot header).
+include apps/calculation/Makefile
```
boricj: The list of Makefiles to include should be stored inside a variable and displayed inside `make info` (like APPS_LIST in #73). Otherwise automatized tools like a build-your-own-image couldn't actually select which apps to build and it would needlessly pollute `git status` too.

EmilieNumworks: You are right. I am though merging this commit as it is a good step forward and make the necessary change in a future commit.

PR #134, opened 2017-09-14
Overlay keyboard image and reposition buttons for the simulator

urashid64:
To make the simulator look more like the actual device.
sim


urashid64:
Looks like there is a Fluid version mismatch.  Will close the PR for now.

adriweb:
This looked pretty great, hopefully you'll be able to make it work with the versions they use!

PR #136, opened 2017-09-15
Fix #108 and #109 storing expressions containing variables

### LINKED ISSUE #108: bug when the result of a formula is stored in a variable used in the formula
1->A returns 1 which is OK
2*A+3->A -> returns 3 instead of 5 and 3 is stored in A
This behavior is as if the variable we store something in is zeroed before the expression is calculated

This bus is in 1.1.0, 1.1.1 and 1.1.2
I don't know if it's already present in 1.0.3

### LINKED ISSUE #109: bug: store ANS into a var
Having a result ans using
`ans->A`
returns "undef" and `A` is empty.

jacobly0:
When m_value was passed to setExpressionForSymbolName directly, and the m_symbol existed in the global context, it ended up evaluating the expression relative to the global context instead of the current context.


Ecco:
Excellent catch. Thank you @jacobly0 !

PR #138, opened 2017-09-15
Improve the [x,n,t] key.

jacobly0:
Inside the first argument of `diff` or `int`, `x,n,t` should type `x`.
Inside the first argument of `product` or `sum`, `x,n,t` should type `n`.
Otherwise, fallback to `x` in the graph app, `n` in the sequence app, and `X` elsewhere since `x` would be useless.


EmilieNumworks:
The code is smart and effective, thank you jacobly0.

PR #140, opened 2017-09-15
[Fix #130] [poincare] Fix recursion bug in adaptiveQuadrature() integration

### LINKED ISSUE #130: Integrating the bell curve returns "undef"
int(e^(-(x^2)),-5,5) returns undef

I'm pretty sure the normal distribution function is well-defined throughout this range.  Doing the integral from -5 to 0, and from 0 to 5, are both successful, as is doing it from -1 to 1.  I'm happy to be shot down, but I think the function ought to be well-defined and thus the integral calculable.

Software version is 1.1.2, built under Fedora 26.

ian-abbott:
The recursive adaptiveQuadrature() function is initially called with the
maximum number of recursion levels to perform, so adaptiveQuadrature()
should decrement numberOfRecursions at each recursion level.

When adaptiveQuadrature() needs to recurse, it splits the interval in
two, calculates the integral for each half by calling itself
recursively, and adds the result from each half.  However, it then
ignores this result and returns NAN because it is checking a stale
quadKG.absoluteError value.  Rearrange the code to avoid that.

### DIFF
--- a/poincare/src/integral.cpp
+++ b/poincare/src/integral.cpp
@@ -171,14 +171,14 @@ template<typename T>
 T Integral::adaptiveQuadrature(T a, T b, T eps, int numberOfIterations, VariableContext<T> xContext, AngleUnit angleUnit) const {
   DetailedResult<T> quadKG = kronrodGaussQuadrature(a, b, xContext, angleUnit);
   T result = quadKG.integral;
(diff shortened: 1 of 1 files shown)


EmilieNumworks:
Good point, indeed. There was definitely something wrong with the previous implementation. Thank you!

Issue #111, opened 2017-09-08
[Suggestion] Shift Alpha Lock changes


ian-abbott:
Currently, the following ShiftAlphaStatus values are supported:
 * `Default` - no modifiers in effect.
 * `Shift` - next key invokes shifted function - title bar shows `shift`.
 * `Alpha` - next key produces lower case character - title bar shows `alpha`.
* `ShiftAlpha` - next key produces upper case character - title bar shows `ALPHA`.
* `AlphaLock` - locked in lower case alpha mode - title bar shows padlock icon + `alpha`.
* `AlphaLockShift` - locked in lower case alpha mode, but next key produces upper case character - title bar shows padlock icon + `alpha`.
* `ShiftAlphaLock` - locked in upper case alpha mode - title bar shows padlock icon + `ALPHA`.

There are some issues with the above:

1. There is no visual distinction between the `AlphaLock` and `AlphaLockShift` states.
2. There is no way to enter a lower case character in the `ShiftAlphaLock` state other than by pressing the 'alpha' key to return to the `Default` state and proceeding from there.
3. Pressing the 'alpha' key in the `AlphaLockShift` state goes back to the `Shift` state. Although reasonably logical, it may be unexpected as there is no visual indication that 'shift' has previously been pressed.

I came up with a couple of possible solutions. One of them involved adding an additional `ShiftAlphaLockShift` state to allow a lower case character to be entered when locked in upper case alpha mode, and to add a couple of new title bar strings `Alpha` and `aLPHA` to distinguish all the current states. However, I think this would make the interface too complicated and confusing.

The simpler solution I came up with is to eliminate the `AlphaLockShift` state altogether and allow the 'shift' key to toggle between the `AlphaLock` and `ShiftAlphaLock` states, so the title bar will keep displaying the padlock icon, but toggle between showing `alpha` and `ALPHA`. This is consistent with the existing use of the 'shift' key to toggle between the `Alpha` and `ShiftAlpha` states.

artaxxx:
Fixed by #141 

Thanks @ian-abbott !

PR #141, opened 2017-09-15
Change shift alpha lock behaviour

ian-abbott:
This is for [Issue #111 - \[Suggestion\] Shift Alpha Lock changes](../issues/111).

I don't know if this is something the project maintainers think is a good idea or not!


artaxxx:
Thanks @ian-abbott for your PR!

I think you found the correct answer to the issue #111 you posted.

The main problem was to produce a lower case character in "ALPHA lock" mode. Your first issue about visual distinction between "alpha" and "Alpha" was also relevant and adding new title bar strings was unnecessarily complex for the end user. So the idea of a toggle is very smart! 

In conclusion, it's a really good idea. Thank you very much!

PR #142, opened 2017-09-15
Graphical Keyboard Overlay for Simulator and matching buttons

urashid64:
This will make the simulator look more like the real device.  Modified the label on a couple of buttons to fit the available space.

Had to account for the differences in FLTK fluid app between the version I was using (1.3.4) and libfltk1.3-dev.  Verified on Travis before creating the PR.

The image below shows the simulator on my Mac desktop.  Hope it is useful.

sim


boricj:
At this point, I'd change the border/background color of the window to blend with the keyboard. The latter looks like a sticker cheaply pasted on otherwise.

adriweb:
It would give something more or less like this: 
test

The color difference on the left and the right is a bit annoying to handle though.

urashid64:
Thanks for the suggestions.  I expanded the image to cover the margins and blended with a soft brush a little bit.  I think it looks better now:

sim2

adriweb:
Looks nice :)

garycmartin:
I'd love to get this pull working on a recent version, anyone know if it's possible to apply? I've not had much luck here, but my git hub foo may be letting me down – any hints appreciated, I was manually trying to pull individual commits to test it and that can't be right ;)

UDXS:
Can this please be merged?

boricj:
@UDXS It'll need to be squashed, cleaned up and rebased first. It should be relatively easy as long as one knows how to properly Git, but unless the original author updates the pull request's original branch a new pull request will be needed.

boricj:
So I went ahead and did it. Since I obviously can't push it on the original branch to update this pull request, it's available at https://github.com/boricj/epsilon/tree/new_skin_simulator instead. I'll give the original author a bit of time to react, if he doesn't then I'll just open my own pull request and ask the NumWorks team to close this one.

urashid64:
@boricj Thanks for fixing the pull request.  I was actually working on a local version to create a new pull request but this seems easier :)

urashid64:
Added fluid as an additional package.  Also, had to re-add undef None to avoid the conflict with X.

PiGuy2:
Is anything still happening with this pull request?

boricj:
At this point, very unlikely. Especially since the SDL-based targets has effectively rendered the FTLK-based simulator obsolete (except for development purposes).

PR #143, opened 2017-09-16
Make the square bracket symbols symmetric and add floor/ceiling symbols

jacobly0:
This displays `floor(x)` as &lfloor;x&rfloor; and `ceil(x)` as &lceil;x&rceil;.  Also unifies the logic for displaying [x], |x|, &lfloor;x&rfloor;, and &lceil;x&rceil; which reveals a visual asymmetry with the original [] that is also corrected.


EmilieNumworks:
Very good factorised code! The result looks great, thank you.

PR #149, opened 2017-09-17
Add orgin/end keyboard shortcuts, by pressing shift left or shift right

zardam:


EmilieNumworks:
That is indeed a good shortcut to add. This behaviour should also be added in TextArea (used in python editor) for more consistency. Thank you!

PR #155, opened 2017-09-20
Use GCC by default on the simulator

Ecco:
Many users don't have clang installed, and those who have very often
also have gcc installed and/or aliased to clang.


adriweb:
IIRC, GCC >= 5.x is required ; for some reason the latest 4.9 (while C++11 compliant), doesn't like a few things (there are only like 3 or 4) in the code...
Some of those things are fixable, and it ends up compiling fine, but there are link errors on template specializations, that I haven't been able to solve...

PR #161, opened 2017-09-24
Add orgin/end keyboard shortcuts, by pressing shift left or shift right to textArea

zardam:


escher/src/text_area.cpp
```diff
@@ -307,8 +313,12 @@ TextArea::TextArea(Responder * parentResponder, char * textBuffer,
 bool TextArea::TextArea::handleEvent(Ion::Events::Event event) {
   if (event == Ion::Events::Left) {
     m_contentView.moveCursorIndex(-1);
+  } else if (event == Ion::Events::Origin) {
+    m_contentView.moveCursorIndexAbsolute(0);
   } else if (event == Ion::Events::Right) {
     m_contentView.moveCursorIndex(1);
+  } else if (event == Ion::Events::End) {
+    m_contentView.moveCursorIndexAbsolute(INT_MAX);
```
EmilieNumworks: m_contentView.moveCursorGeo(-INT_MAX, 0);

EmilieNumworks: m_contentView.moveCursorGeo(INT_MAX, 0);

escher/src/text_area.cpp
```diff
@@ -307,8 +313,12 @@ TextArea::TextArea(Responder * parentResponder, char * textBuffer,
 bool TextArea::TextArea::handleEvent(Ion::Events::Event event) {
   if (event == Ion::Events::Left) {
     m_contentView.moveCursorIndex(-1);
+  } else if (event == Ion::Events::Origin) {
+    m_contentView.moveCursorIndexAbsolute(0);
```
EmilieNumworks: m_contentView.moveCursorGeo(-INT_MAX, 0);

escher/src/text_area.cpp
EmilieNumworks: If you do that, you will have to correct the following method:
size_t TextArea::Text::indexAtPosition(Position p) {
  if (p.line() < 0) {
    return 0;
  }
  int y = 0;
  const char * endOfLastLine = nullptr;
  for (Line l : *this) {
    if (p.line() == y) {
      size_t x = max(p.column(),  0);
      x = min(clippedPColumn, l.length());
      return l.text() - m_buffer + x;
    }
    endOfLastLine = l.text() + l.length();
    y++;
  }
  assert(endOfLastLine != nullptr && endOfLastLine >= m_buffer);
  return endOfLastLine - m_buffer;
}

EmilieNumworks:
Thanks zardam, the new feature works perfectly! However, I would avoid adding a new methods to TextArea::ContentView. I think, you could use moveCursorGeo (moveCursorIndexAbsolute is just a special case of this one).
Doing so, you might find a bug in TextArea::Text::indexAtPosition(Position p) which should clipped the position p within the text area (right now, it clips the position on the left but not on the right ...)

EmilieNumworks:
I finally made the change in #166, thank you for your help zardam.

Issue #133, opened 2017-09-14
Add conversion from/to bin, oct, dec, hex


Zezombye:
For engineering, these conversions could be useful. Also allow binary operators (and, or, not, xor, xnor...).

adriweb:
Or rather, arbitrary base conversion. But some useful shortcuts like other calculators have.
For instance with a base prefix + conversion suffix/helper: `0b0101110▶Base16` on the Nspire, but it could also be `toHex(...)`.

ian-abbott:
This is as good a place as any to discuss possible input and display formats for hexadecimal, octal, and binary numbers. A prefix starting with a `0`, such as `0h` for hexadecimal might be a bit ambiguous, so how about `:h` or `?h` as a hexadecimal prefix?

boricj:
Given poincare's available symbols in its lexer and the NumWorks keyboard, we could use the format `xxxx_yy`, where `100_16` = `100_h` = 256 and pretty-print it to ![equation](http://www.sciweavers.org/tex2img.php?eq=100_%7B16%7D&bc=White&fc=Black&im=png&fs=10&ff=arev&edit=0). Then, typing `ans_2` would print ![equation](http://www.sciweavers.org/tex2img.php?eq=100000000_2&bc=White&fc=Black&im=png&fs=10&ff=arev&edit=0). Like angles, a radix unit in the settings app would specify the default radix.

ian-abbott:
@boricj There would be more back-tracking in the input parsing that way. In order to prevent `10A2_16` being parsed as `10*A*2_16` (with implicit multiplication), there would need to be greedy backtracking to reparse the previous hex digits that have already been parsed as something else.

boricj:
Not unless hex input is restricted to lowercase. Which would be easier to input anyways on the real calculator.

RedGl0w:
This issue can be closed since https://github.com/numworks/epsilon/commit/7c2ef5847b81141525320880c289f893a598491f

boricj:
@RedGl0w Octal is missing IRRC.

Ecco:
@boricj So is base 47 😄 

I'm going to close this issue since most of it has been addressed. Feel free to open new issues if needed .

debrouxl:
Well, handling octal in upstream would be easy and take little space, given that the framework for base conversion is already fully operational ;)
[...]

PR #163, opened 2017-09-24
Numerical bases support

boricj:
This fixes in part #133 by providing a mechanism to input constant numbers in hexadecimal, octal and binary (`hex(n)`, `oct(n)` and `bin(n)`).

These pseudo-functions are handled inside the lexer and they can't be used to output numbers in other bases. Implementing that will require deeper modifications to the code (I tried, but I got so confused I gave up...).

Translations in languages other than English and French are missing.

### DIFF
--- a/poincare/Makefile
+++ b/poincare/Makefile
@@ -58,6 +58,7 @@ objs += $(addprefix poincare/src/,\
   matrix.o\
   multiplication.o\
   naperian_logarithm.o\
+  to_base.o\
   nth_root.o\
   opposite.o\
   parenthesis.o\
@@ -142,6 +143,7 @@ tests += $(addprefix poincare/test/,\
   subtraction.cpp\
   symbol.cpp\
   store.cpp\
+  to_base.cpp\
   trigo.cpp\
   vertical_offset_layout.cpp\
 )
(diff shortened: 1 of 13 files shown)


boricj:
I rebased the pull request to adjust for the new i18n stuff. Travis CI fails because translations other than French and English are missing, but it builds and works fine with `EPSILON_I18N_LANGUAGES='en fr'`

boricj:
So I've resurrected this pull request and implemented all numerical base functionalities from base 2 to base 36. To keep things simple, converting from a base works by (ab)using the lexer and converting to a base hooks itself inside the simplification process. Syntax might be a bit on the heavy side in the more complex cases (`tobase(frombase(10,7),4)`), but it does not introduce new notations from the point of view of the user.

Note: translations other than French use English since I'm not a polyglot.

boricj:
Well, 2D edition mode doesn't quite like my numerical bases. Everything still works fine in 1D edition mode though, so I'm not _too_ far off target.

adriweb:
In 2D, a layout for bases could be some sort of subscript on the right (pretty much how they're represented anywhere, AFAIK)

boricj:
I ditched pretty print, but it turns out `#` is not on the NumWorks keyboard... I picked `:` as the least worst choice available, so typing `6*7→:13` yields `42:13 = 54`. On one hand this notation massively simplifies usage and typing, on the other it's _really ugly_ with `:`... Maybe syntax highlighting and/or another character would make it prettier? Still open to suggestions.

Oh, and the CI is broken. It says it can't find the `gcc-arm-embedded` package.

boricj:
Pull request thoroughly bitrotten, closing.

PR #170, opened 2017-09-27
[kandisky] Generated font-related source data should be const.

adriweb:
This saves up to 30 KB in RAM, as it's not in the same section anymore.

(thanks to @debrouxl)


Ecco:
I can't believe we overlooked this. I would have swore font bitmap data was already in rodata…

Thanks a lot @adriweb and @debrouxl for pointing this out!

PR #172, opened 2017-09-27
Correct wrong output of python complex object.

zardam:
For example "print(complex(1,1))" was "(%s1+1j)" instead of "(1+1j)"


Ecco:
Hi @zardam ! Thank you very much for this PR. Unfortunately we're not going to merge it because it's not relevant anymore in the new Python app. The good news is, the bug is fixed anyway :-)

PR #174, opened 2017-09-28
Split the i18n data in multiple files

Ecco:
A Python script aggregates all i18n files and merges them.
We now can pick a list of locales to build in EPSILON_I18N_LANGUAGES.


apps/i18n.py
```diff
@@ -0,0 +1,178 @@
...
+    f.write("namespace I18n {\n")
+    f.write("\n")
+    for message in data["messages"]:
+        for locale in locales:
+            if not locale in data["data"]:
+                sys.stderr.write("Error: Undefined locale \"" + locale + "\"\n")
+                sys.exit(-1)
+            if not message in data["data"][locale]:
+                sys.stderr.write("Error: Undefined key \"" + message + "\" for locale \"" + locale + "\"\n")
+                sys.exit(-1)
```
boricj: That's a bit extreme for a missing translation. Third-party contributors/app developers like me typically aren't fluent in all of Dutch, English, Spanish, French and Portuguese (and probably more in the future).

If an English translation of a missing key is available, I'd output a warning and use it. At least Travis won't fail our pull requests just because we aren't polyglots and we can actually see what's truly missing, instead of blindly padding with junk to keep the compiler happy (like in #163).

apps/i18n.py
```diff
@@ -0,0 +1,178 @@
...
+            f.write("constexpr static char " + locale + message + "[] = " + data["data"][locale][message] + ";\n")
+    f.write("\n")
+    f.write("constexpr static const char * messages[%d][%d] = {\n" % (len(data["messages"]), len(locales)))
+    for message in data["messages"]:
+        f.write("  {")
+        for locale in locales:
+            if not locale in data["data"]:
```
boricj: Duplicated and redundant tests (the first `for` loop already cover these cases).

apps/i18n.py
Ecco: At the moment we want to support all 5 built-in languages (en,es,fr,de,pt) so we wouldn't merge a feature without the appropriate translation anyway.

That being said, you make a very good point about contributors not being fluent in all those languages. That's not a big issue though: we'll help :)

Ecco:
Thanks for the review @boricj ! We'll enforce the 5 languages requirement for now, but like I said, we'll help contributors with translations if needed.

PR #181, opened 2017-10-01
[settings] allow left event to pop out of the lang controller (fix #180)

### LINKED ISSUE #180: [bug] Can't leave Language submenu in the Settings app with left arrow
Any other submenu in the `Settings` app can be left with both the left arrow and back key.

adriweb:
The condition check is done last since the event could be handled by the deeper implementation (shared language controller) already.


EmilieNumworks:
Thank you for fixing this !

PR #189, opened 2017-10-04
Source code spelling fix reel => real

ian-abbott:
Some source code spelling corrections for "real" misspelled as "reel". No changes to functionality.


Ecco:
Good catch. Thank you @ian-abbott !

PR #190, opened 2017-10-04
i18n Occitan translation

JeanGauthier:
I hope I changed all the required files, if not tell me what to change.
Special thanks to Danís for the review


adriweb:
Looks like a squash would be quite useful here :)

JeanGauthier:
@adriweb is this some kind of test?

adriweb:
Nope, it's the word meaning that all commits will be "squashed together", so in the end you'd end up with one (well, you choose what you squash) instead of a whole lot.
This allows a much cleaner history before it's pushed, so commits are grouped thematically for instance.

But it doesn't matter that much since the maintainers can squash-merge automatically.

JeanGauthier:
Thanks, I think I get it. If it'll be done automatically it's fine then :)

JeanGauthier:
Hello @Ecco @EmilieNumworks any news for merge?
Best regards

EmilieNumworks:
Hi @JeanGauthier, we won't merge Occitan translation. The number of users who requested it is not worth its maintenance. Feel free to fork our repository!
Thanks again!

Issue #115, opened 2017-09-09
isnan and isinf belongs to std namespace


Mojo4242:
Hi !

I'm discovering epsilon and for my first build, I used a docker container with the latest Ubuntu:
```
apt-get update
apt upgrade
apt install git make gcc libfreetype6-dev bison g++ flex python clang libfltk1.3-dev
git clone https://github.com/numworks/epsilon.git
cd epsilon/
make PLATFORM=simulator
```
I got plenty of error in a lot of files like:
```
poincare/src/complex.cpp:362:7: error: use of undeclared identifier 'isinf'
  if (isinf(truncatedMantissa) || isnan(truncatedMantissa)) {
      ^
poincare/src/complex.cpp:368:9: error: use of undeclared identifier 'isnan'
    if (isnan(newLogBase10) || isinf(newLogBase10)) {
        ^
```
I suppose this error didn't showed to other people before and I don't quite understand why.
```
g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/5/lto-wrapper
Target: x86_64-linux-gnu
[...]

PR #191, opened 2017-10-05
Fix simulator compile with gcc.

lyda:
Building with `make PLATFORM=simulator CXX=g++ CC=gcc LD=g++`
would fail with errors about isnan and isinf missing.  They're
in the std namespace.

So added the relevant "using" lines for each function.

(Note: this would be an alternative to some part of #115)

### DIFF
--- a/apps/probability/calculation_controller.cpp
+++ b/apps/probability/calculation_controller.cpp
@@ -11,6 +11,8 @@
 
 using namespace Poincare;
 using namespace Shared;
+using std::isinf;
+using std::isnan;
 
 namespace Probability {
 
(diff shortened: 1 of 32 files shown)


adriweb:
Yep, it is a known issue.
However your modifications can't be done this way (see the travis output). It looks like https://github.com/numworks/epsilon/pull/123 is more complete, though, and the travis builds is OK.

You may know how to do that already, but just in case, you can try the PR yourself by downloading [https://patch-diff.githubusercontent.com/raw/numworks/epsilon/pull/123.patch](the patch file) and applying it locally on your repo. It should then build fine.

Ecco:
Hi @lyda ! Thank you for your pull request. Unfortunately it somehow duplicates the functionalities of #123 which we just merged (after a thorough review we thought #123 was a more comprehensive patch). The bad news is that we won't be able to merge your PR. The good news is, the problem is fixed :-)

Issue #203, opened 2017-10-10
epsilon-build-related python script issue: python 3.6


yeKcim:
Can't compile with python 3.6. First, there is the sets  #197, but if I correct this bug, there is these : 

$ make
```
…
I18N    apps/i18n.cpp
Traceback (most recent call last):
  File "apps/i18n.py", line 167, in <module>
    data = parse_files(args.files)
  File "apps/i18n.py", line 55, in parse_files
    name,definition = split_line(line)
  File "apps/i18n.py", line 40, in split_line
    return (match.group(1), source_definition(match.group(2)))
  File "apps/i18n.py", line 31, in source_definition
    ion_characters.append(ion_char(i18n_letter))
  File "apps/i18n.py", line 26, in ion_char
    return "'" + normalized + "'"
TypeError: must be str, not bytes
make: *** [apps/Makefile:81: apps/i18n.cpp] Error 1
```

$ python --version
```
Python 3.6.2
```

Seems to be little bug, it could be interesting to be 3.6 compliant…

ian-abbott:
I don't think it works with any version of Python 3. The easiest fix is to run python2 explicitly, but that assumes the system adheres to [PEP 394](https://www.python.org/dev/peps/pep-0394/).

yeKcim:
add 2 at the end of the first line `#!/usr/bin/python2` in apps/i18n.cpp is not efficient

ian-abbott:
@yeKcim I think you mean apps/i18n.py and "sufficient". The script is not executed directly via the magic "shebang" line (`#!`). It is run as an argument of the python command from apps/Makefile. So apps/Makefile needs a small change to run python2 instead of python.

yeKcim:
I tried
```
export PYTHON=python2.7
make
```

Still have errors.

Thank you for correcting my English @ian-abbott (and file name).

adriweb:
Well sure, it's just calling `python` directly, not an env var. https://github.com/numworks/epsilon/blob/e6ed6c7/apps/Makefile#L81

yeKcim:
the only solution I found: https://wiki.archlinux.org/index.php/Python#Python_2

Add a /usr/local/bin/python file:
```
#!/bin/bash
script=$(readlink -f -- "$1")
case "$script" in (~/epsilon/*)
    exec python2 "$@"
    ;;
esac

exec python3 "$@"
```
(change `~/epsilon/` with your git dir)

With this file, manjaro or archlinux can compile epsilon

[...]

PR #205, opened 2017-10-11
Workaround for issue #203

ian-abbott:
Work around the Python 3 incompatibilities by running the apps/i18n.py script with the python2 executable.

### DIFF
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -78,7 +78,7 @@ i18n_files += $(addprefix apps/,\
 apps/i18n.h: apps/i18n.cpp
 apps/i18n.cpp: $(i18n_files)
 	@echo "I18N    $@"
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Don't know if it helps but we had to deal with a similar issue in Node.js and since we can't assume the system has a python2 binary, I addressed it like this: nodejs/node@c6da5c8cdfae9e80d63f090c4289fd146c3175df

(I readily acknowledge it's both a hack and black magic but it works.)

Ecco:
@bnoordhuis That is indeed interesting! That being said, our Python scripts are simple enough so we can make them work on both Python 2 and Python 3.

Ecco:
Hi @ian-abbott ! Sorry for the delay replying to this PR. This does indeed fix things for people whose `python` is in fact Python 3. But this would also break systems where Python 2 is not aliased as `python2` (e.g. macOS).

I'm going to close this PR since it is superseeded by #361 

Anyway, thank you for your help on this!

PR #208, opened 2017-10-14
Corrects wrong RCC QSPIEN register definition

zardam:


Ecco:
Good catch. Thank you @zardam !

PR #210, opened 2017-10-16
Avoid using clang by default

Ecco:


Ecco:
> Btw, m_array{"..."} (as opposed to m_array("...")) has worked since gcc 4.8.1.

Indeed, thanks for pointing this out. But anyway, Ubuntu 14.04 ships with gcc-4.4 and 16.04 with gcc-4.7…

Not that I enjoy working around compiler bugs, but we may as well fix it once and for all…

PR #212, opened 2017-10-18
Switching from horizontal to vertical scrolling on the main menu

boricj:
Once again, it's proof-of-concept time.

The first commit simply adds the variable `EPSILON_APPS` that contains the list of apps to include. It's the bare minimum so that I can make my own apps without polluting `apps/Makefile`.

The second commit switches the main menu from horizontal to vertical scrolling. There are a bunch of reasons for doing that:
- Going above 8 apps makes the home menu go berserk and I couldn't be bothered to fix it ;
- Scrolling down brings 3 new icons whereas scrolling right brings only 2, so it should be more wieldy with large numbers of apps ;
- It acts like a file explorer window and thus the order of apps follows intuitively what's declared inside `EPSILON_APPS`.

I can see a couple of drawbacks:
- Less horizontal space for app labels ;
- Up to 2 empty spaces with a lone icon on the bottom row ;
- We do lose somewhat a bit on the alignment combo (for lack of better words) ;
- Less PlayStation 4-like, more Casio-like ;
- NumWorks would have to touch up a bunch of artwork ;
- Currently, having 6 or less apps does not reclaim the scroll bar space.

I don't actually _expect_ this to be merged, but it's a neat trick nonetheless.

### DIFF
--- a/apps/home/app_cell.cpp
+++ b/apps/home/app_cell.cpp
@@ -13,7 +13,7 @@ AppCell::AppCell() :
 
 void AppCell::drawRect(KDContext * ctx, KDRect rect) const {
   KDSize nameSize = m_nameView.minimalSizeForOptimalDisplay();
-  ctx->fillRect(KDRect(0,  bounds().height()-nameSize.height() - 2*k_nameHeightMargin, bounds().width(), nameSize.height()+2*k_nameHeightMargin), KDColorWhite);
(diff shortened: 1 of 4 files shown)


Ecco:
Hi @boricj, thank you very much for this. We really appreciate that you went all the way up to an actual working demo. We need to try this out and show it to as many people as we can to get their feedback on that interface. Thanks again!

boricj:
Closing since commit 197027c9 made the switch.

PR #215, opened 2017-10-20
Use Calculus instead of Calculations in the toolbox

AlainBusser:
Confidence intervals shall be computed using the [1.96 ratio](https://en.wikipedia.org/wiki/1.96). 

Original French report below.

> Tant au bac qu'au BTS les intervalles de confiance pour une proportion se calculent avec une demi-largeur égale à 1,96 fois l'écart-type et non avec la formule simplifiée vue en 2nde (et qui n'est plus au programme).
> 
> 
> **Par ailleurs**, le terme regroupant les calculs différentiel et intégral est [calculus](https://en.wikipedia.org/wiki/Calculus) et non calculation (en anglais); en français c'est [analyse](https://fr.wikipedia.org/wiki/Analyse_(math%C3%A9matiques))


adriweb:
You'll want to add `nbproject/` and `output.map` to your .gitignore (and not commit that) ;)
Also, it'd be nicer to have a distinct commit for the "Calculus" name fix.

Now more to the point: what about in other countries? It might be subject to the same kind of discussions regarding the definitions of quartiles...

Ecco:
Thank you very much for your feedback @AlainBusser, it's very helpful. I took the liberty of translating your bug report in English since some of our contributors don't speak French.

AlainBusser:
oops I did only check with the simulator and did not see that the erf function is not that standard (seems to depend on the C++ version); so I programmed it with a formula from Abramowitz and Stegun

matheod:
In ES it's with the 2nd formula.

AlainBusser:
read it [on reddit](https://www.reddit.com/r/numworks/comments/98osui/is_it_worth_is_to_buy_the_numworks_calculator/) : When you compute a limit, a differential or an integral, you are doing calculus, not calculation :wink:

EmilieNumworks:
Hi @AlainBusser, "Calculus" or "Analyse" are a Math branch but our "Calculation" app can't pretend to be a "Calculus" app. It only has an interface to compute calculations (it has nothing to explore limits, asymptotes or whatsoever).

PR #224, opened 2017-10-24
Coverity: fix unitialized  pointer fields, out of bounds reads ...

EmilieNumworks:
Change-Id: Iae0656ffe667692dfb3bebf9bb7913b5872175c6


poincare/src/least_common_multiple.cpp
```diff
@@ -49,7 +49,7 @@ Evaluation<T> * LeastCommonMultiple::templatedEvaluate(Context& context, AngleUn
     a = b;
     b = r;
   }
-  return new Complex<T>(Complex<T>::Float(std::round((T)(product/a))));
+  return new Complex<T>(Complex<T>::Float(product/a));
```
adriweb: Is this what's causing the travis blackbox build to fail?

EmilieNumworks: No, it was a wrong assertion I added in poincare/src/integer.cpp ... Thank you for helping with coverity!

adriweb: Ah, right. And no problem, Coverity is a very useful tool I've used on other projects, so I'm glad you find it helpful too :)

PR #225, opened 2017-10-24
Implement 'auto-ans' for Minus

Grahack:
This one was obvious, I can do the same for other operations (ln log sqrt…).

### DIFF
--- a/apps/calculation/text_field.cpp
+++ b/apps/calculation/text_field.cpp
@@ -18,7 +18,8 @@ bool TextField::handleEvent(Ion::Events::Event event) {
     return true;
   }
   if (textLength() == 0 &&
-      (event == Ion::Events::Multiplication ||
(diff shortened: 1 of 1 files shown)


Grahack:
This was a bad idea since with my code it becomes impossible to type negative numbers.
Sorry for the noise.

Ecco:
Hi @Grahack ! Thank you for your contribution. It turns out we already had a lenghty discussion on this matter (see #82). We decided to explicitly remove auto-insertion of ans for subtractions, and so far we got some rather positive feedback on the current state of things. I'll close this PR for now, feel free to discuss on #82 if you don't like the current solution :-)

Grahack:
Just for the record, my idea about `ln`, `log`, `sqrt`, `exp`… was bad for the same reason.

PR #233, opened 2017-11-01
A few misc. coverity fixes

adriweb:
Those are quite minor, so the fixes are trivial, but other (more important) issues exist (for instance [that one](https://scan4.coverity.com/reports.htm#v15949/p12091/fileInstanceId=28662377&defectInstanceId=5676388&mergedDefectId=200502), buffer overflow) but I haven't really looked into them yet.
There was a false-positive, too, involving an out-of-bounds access regarding w_n, but that's fine for now since it's not in use yet (and when it will be, the vector will be correctly sized).


adriweb:
(rebased)

Ecco:
Thank you for rebasing!

PR #238, opened 2017-11-11
[Windows] Let's make the compiler work

gpotter2:
Trying to make `make` succeed on Windows.

**New doc:**
```
pacman -S mingw-w64-x86_64-gcc mingw-w64-x86_64-freetype mingw-w64-x86_64-fltk mingw-w64-x86_64-cmake git make bison python2
ln -s /usr/bin/python2.exe /usr/bin/python.exe
export PATH=/mingw64/bin:$PATH
```

Previous commands:
- fixed python 2 (python = python 3)
- missing depedencies (g++)

The patch:
- fixes "\r\n" being wrongly parsed
- adds exe to .gitignore + force LF
- detects MINGW when using MSYS2

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # No objects files.
 *.o
 *.elf
+*.exe
 
 # No dependency files
 *.d
--- a/apps/i18n.py
+++ b/apps/i18n.py
@@ -53,7 +53,7 @@ def parse_files(files):
             data[locale] = {}
         with open(path, "r") as file:
             for line in file:
-                name,definition = split_line(line)
+                name,definition = split_line(line.strip())
                 if locale == "universal":
                     if name in messages:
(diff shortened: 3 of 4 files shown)


gpotter2:
Current status: Not working.
```
CXX     apps/probability/app.o
In file included from apps/probability/law_controller.h:8:0,
                 from apps/probability/app.h:6,
                 from apps/probability/app.cpp:1:
libaxx/include/new:6:53: error: 'operator new' takes type 'size_t' ('long long nsigned int') as first parameter [-fpermissive]
 inline void * operator new(size_t size, void * ptr) noexcept {
                                                     ^~~~~~~~
make: *** [Makefile:118: apps/probability/app.o] Error 1
```

adriweb:
So you're trying to build for PLATFORM=device, from windows, with mingw?
I'm not sure if it's possible as-is. At least, your changes don't seem to be enough since you'll need all the arm toolchain things.

For a simulator build, your changes might be enough, though (maybe the msys check to put in the simulator file as well)

adriweb:
( See #239 )

PR #243, opened 2017-11-16
Enable instruction and data cache

zardam:
Enabling instruction and data cache speed up the calculator a little bit.

For example, the mandelbrot set in the python application takes 1 minutes 30 seconds to render with this modification, and about 2 minutes without.

This should not raise power consumption  as stated [here](http://www.st.com/content/ccc/resource/technical/document/application_note/13/0a/06/b9/1e/2f/4d/9d/DM00096220.pdf/files/DM00096220.pdf/jcr:content/translations/en.DM00096220.pdf) (but not clearly) or [here](http://www.st.com/content/ccc/resource/technical/document/application_note/77/74/4f/da/9b/9c/43/e7/DM00033348.pdf/files/DM00033348.pdf/jcr:content/translations/en.DM00033348.pdf) (for the F2 family) in the section ART accelerator.

Instruction prefetch may be also enabled, but this can make the power consumption raise by 10%

### DIFF
--- a/ion/src/device/device.cpp
+++ b/ion/src/device/device.cpp
@@ -164,6 +164,10 @@ void initClocks() {
    * The spec tells us that at 2.8V and over 90MHz the flash expects 3 WS. */
   FLASH.ACR()->setLATENCY(3);
 
+  /* Set flash instruction and data cache */
+  FLASH.ACR()->setDCEN(true);
+  FLASH.ACR()->setICEN(true);
+
   /* We're using the high-speed internal oscillator as a clock source. It runs
    * at a fixed 16 MHz frequency, but by piping it through the PLL we can derive
    * faster oscillations. Combining default values and a PLLQ of 4 can provide
@@ -229,4 +233,3 @@ void shutdownClocks() {
 
 }
 }
-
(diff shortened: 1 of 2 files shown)


adriweb:
Nice!

Ecco:
That is an excellent pull request. We definitely should look into getting the most out of the ART.
I'm not merging this right away because I'd like to investigate a couple things before:
* First, check there is indeed no power consumption difference
* Second, investigate the actual power consumption when prefetching is enabled.

Anyway, thank you very much @zardam for this PR!

EmilieNumworks:
After some tests on power consumption:

|                        | With screen | Without screen | Off | 
| ------------- | ------------- |------------- |------------- |
| No ART configuration  |  ~112 mA | ~30 mA | ~29 uA |
| Instruction & data cache  |  ~108 mA | ~25 mA | ~29 uA |
| Prefetching  |  ~115 mA | ~30 mA | 29 uA |
| Prefetching+Instruction & data cache |  ~115 mA | ~30 mA | 29 uA |

I tested performance on Mandelbrot(10):

|                        | Time for execution | 
| ------------- | ------------- |
| No ART configuration  |  1'11'' |
| Instruction & data cache  |  56'' |
| Prefetching  |  1'03'' |
| Prefetching+Instruction & data cache |  53'' |

Indeed, cache does not seem to raise the power consumption. Thank you @zardam for the commit. 
Plus, the increase of power consumption  when fetching instructions is negligible compared to the screen consumption. I am thus enabling prefetching as well!

PR #251, opened 2017-12-05
Fix: invalid characters on windows simulator #240

### LINKED ISSUE #240: invalid characters on windows simulator
I compiled linux and windows simulator as explained here : https://www.numworks.com/resources/engineering/software/sdk/
No problem with linux, but the windows version displays squares on some keys.

linux : linux

windows : windows

ydupas:
Hi all,

A solution to fix #240, is to set font "Arial Unicode MS" for Windows platform. It's because, default Windows Arial is not fully unicode...

ion/src/simulator/init.cpp
Fl::set_font(FL_SYMBOL, "Arial Unicode MS");

ion/src/simulator/keyboard/fltkkbd.cpp
m_buttons[k]->labelfont(FL_SYMBOL);

### DIFF
--- a/ion/src/simulator/init.cpp
+++ b/ion/src/simulator/init.cpp
@@ -29,6 +29,10 @@ void init_platform() {
 
   Fl_Window * window = new Fl_Window(screen_width+2*margin, margin+screen_height+margin+keyboard_height+margin);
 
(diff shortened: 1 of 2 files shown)


Ecco:
Thank you very much @ydupas !

PR #271, opened 2017-12-08
[poincare] print infinite complex

EmilieNumworks:


AlainBusser:
What if the complex is actually a real number? Should not exp(-inf) be 0.0 and exp(+inf) be inf? I used to use the online simulator to teach the notion of limits and nowadays it is less easy because I need to program in Python with things like

   ```
 from math import *
 oo = float("inf")
 print(exp(-oo))
 print(exp(oo))
```

Issue #267, opened 2017-12-07
windows simulator: Assert error apps/i18n.cpp:2437 and Toolbox title "Zoom :"


ydupas:
Hello,

When I run simulator (1da26f4f65a69075dab2ba9c5ff1d12d5936ae6c) on MSYS2, I have an assert error apps/i18n.cpp:2437, the assert is:
assert(languageIndex > 0);
![assertion failed 1 2 0](https://user-images.githubusercontent.com/34285316/33728080-93330294-db79-11e7-8211-df7fa20122f3.PNG)
Click "Ignore"
And the title of toolbox is "Zoom :"
![toolbox title 1 2 0](https://user-images.githubusercontent.com/34285316/33728119-b38e39f0-db79-11e7-8fe5-0f93afc56980.PNG)

I find a something at apps/math_toolbox.cpp:101 when I
replace: MathToolbox::MathToolbox() : Toolbox(nullptr, I18n::translate(rootModel()->label()))
by: MathToolbox::MathToolbox() : Toolbox(nullptr, "[[[TOTO]]]")
With this change, no assert error and the Toolbox title is "[[[TOTO]]]".
![toolbox title 1 2 0 try find assert error](https://user-images.githubusercontent.com/34285316/33728185-ed3cac2c-db79-11e7-8871-db764c01b7f2.PNG)

Something wrong in or before I18n::translate(rootModel()->label()) at init of MathToolbox

With 1.1.3, it's ok:
![toolbox title 1 1 3](https://user-images.githubusercontent.com/34285316/33728274-337fd902-db7a-11e7-88ba-8839b0d39327.PNG)

A possible commit who introduce this assert error is : 68e237702b5209ebc43aaecc50ef0874f3d54843

PR #275, opened 2017-12-08
Fix: windows simulator: Assert error apps/i18n.cpp:2437 and Toolbox title "Zoom :" #267

ydupas:
Global pref isn't init when the init of MathToolbox:
static GlobalPreferences s_globalPreferences doesn't init on windows platform, move to
static methode sharedGlobalPreferences()

### DIFF
--- a/apps/global_preferences.cpp
+++ b/apps/global_preferences.cpp
@@ -1,6 +1,8 @@
 #include "global_preferences.h"
 
+#if !defined(_WIN32) && !defined(_WIN64)
 static GlobalPreferences s_globalPreferences;
+#endif
 
 GlobalPreferences::GlobalPreferences() :
   m_language(I18n::Language::EN),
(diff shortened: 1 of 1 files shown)


adriweb:
Maybe forcing the initialization would be better.

Something like : `static GlobalPreferences s_globalPreferences{};`
Just an idea, that's untested :)

ydupas:
Yes, but `static GlobalPreferences s_globalPreferences{};` doesn't work.
Actually, I doesn't find other solution, but if you have, I get it!
I doesn't test on device, but it could be move too. And we remove conditional compilation.

Ecco:
Hi @ydupas , thank you very much for this fix. Indeed, removing the conditional compilation seems like a good idea. I wonder why the problem happens only on Windows though. Is this a case of [static init order fiasco](https://isocpp.org/wiki/faq/ctors#static-init-order)?

jacobly0:
Having a function return a local static instead of having a global singleton is a common pattern used to avoid the static initialization order fiasco because local statics are not initialized until the function is called the first time, so this should indeed be done unconditionally.

Ecco:
Yes, of course. That being said, is that what's causing the problem here?

ydupas:
Yes, certainly, we get assert error before reaching the first instruction of main function in ion/src/simulator/boot/main.cpp, line 4

ydupas:
I remove conditional compilation and my test on device is ok. Title toolbox is ok.

Ecco:
That was a great fix. Thank you very much @ydupas!

PR #277, opened 2017-12-08
Cleanup math.h and update tests and cmath.

jacobly0:
Cleans up a duplicate, order, naming, and adds functions recently added to math.h but not cmath.


AlainBusser:
If this allows to compute the probability of a normal variable with more precision (using erf) it is important !

EmilieNumworks:
Hi @jacobly0, I just rebased your PR to be able to merge it. I close this one as it is outdated. Thanks!

PR #285, opened 2017-12-14
[apps] In code: add single quote (') in the catalog

EmilieNumworks:


ydupas:
The single quotes are used by two, 
To avoid returning to the catalog for the second single quote.
It's possible to set two single with Python catalog and place cursor in the middle?

ydupas:
It's good on device to edit script.

Issue #286, opened 2017-12-16
unexpected reset


LRion:
Bug description :

Go to Settings -> Result format
Press the "down arrow" two times, you are now on "Significant figures"
Press the "back button" (= button at the top right of the device)

On my device, this procedure reset all the memory.
This bug gives possibility to exit exam mode without usb connection.

Software version 1.2.0

adriweb:
I it actually present in the simulator, at least when you build in debug: 
```cpp
Assertion failed: (result->m_magic == App::Magic), function app, file escher/src/responder.cpp, line 86.
[1]    11550 abort      ./app.elf
```

**Stacktrace**: 
```
  * frame #0: 0x000000010f908ca4 app.elf`Settings::SubController::handleEvent(this=0x000062300004a040, event=(m_id = '\x05')) at sub_controller.cpp:71
    frame #1: 0x000000010f8b3d39 app.elf`App::processEvent(this=0x0000623000049900, event=(m_id = '\x05')) at app.cpp:49
    frame #2: 0x000000010f8b6775 app.elf`Container::dispatchEvent(this=0x000000010f98a770, event=(m_id = '\x05')) at container.cpp:39
    frame #3: 0x000000010f930d47 app.elf`AppsContainer::dispatchEvent(this=0x000000010f98a770, event=(m_id = '\x05')) at apps_container.cpp:83
    frame #4: 0x000000010f8bda02 app.elf`RunLoop::step(this=0x000000010f98a770) at run_loop.cpp:69
    frame #5: 0x000000010f8bd80d app.elf`RunLoop::run(this=0x000000010f98a770) at run_loop.cpp:24
    frame #6: 0x000000010f8b67f0 app.elf`Container::run(this=0x000000010f98a770) at container.cpp:48
    frame #7: 0x000000010f9310c8 app.elf`AppsContainer::run(this=0x000000010f98a770) at apps_container.cpp:135
    frame #8: 0x000000010f9342e0 app.elf`ion_app() at main.cpp:6
    frame #9: 0x000000010f8237a9 app.elf`main(argc=1, argv=0x00007ffee03dd7c0) at main.cpp:6
```

LRion:
My apologies :
The bug is also present on simulator if I compile the simulator on my desktop.
(The bug is not present on the *online* simulator : https://www.numworks.com/simulator/)

This bug gives possibility to exit exam mode without usb connection.

PR #287, opened 2017-12-18
[apps] Fix bug in settings: do not set the sub controller table as first

EmilieNumworks:


ydupas:
It's fix the issue unexpected reset #286.

PR #292, opened 2017-12-22
Add option USE_LIBGCC to link agains libgcc instead of liba softfloat implementation

zardam:
This leads to smaller binary size:
 - 631 328 bytes when using liba
 - 628 608 bytes when using libgcc

And also to improved performance, at least in the mandelbrot python script. When using mandelbrot(10), it takes about:
 - 54 seconds with liba
 - 42 seconds with libgcc

There is normally no license problem to link against libgcc : https://www.gnu.org/licenses/gcc-exception-3.1-faq.en.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
 matrix:
   include:
     - env: PLATFORM=device EXT=elf
+    - env: PLATFORM=device EXT=elf USE_LIBGCC=1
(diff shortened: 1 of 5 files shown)


adriweb:
That's quite an interesting performance boost.... Nice work :D

Ecco:
Moved to #630

PR #296, opened 2017-12-29
Very small feature : In Home application, Home or Back button select Calculation

maxBzh:
Some very minor change, almost futile, to the Home application.

Like in many handheld devices, when in the Home application, if you push the Home (or Back) button, you will select the first page/application.


Ecco:
Very good idea! Thank you @maxBzh

PR #311, opened 2018-01-15
Clock

boricj:
Yet more proof-of-concept time!

Title says it all. Works both on the simulator and on the real device.

Only English and French translations provided. No 12-hour clock style. No date support whatsoever. Precision of time-keeping potentially poor. Code quality questionable.

But at least my Numworks calculator can show me now how late I am for dinner tonight... >_<


Ecco:
> But at least my Numworks calculator can show me now how late I am for dinner tonight... >_<

😄

boricj:
I don't have the means to precisely quantify the accuracy of my clock, but my calculator is about 13 minutes off after 13 hours, so it's about 1/60 (or 1.67%) faster than real-time in my case.

While that sounds _really_ bad, I did not actually calibrate the LSI RC oscillator at all (the RTC is configured for a static 32 kHz signal). What could be a bigger problem is clock stability, especially since no calibration is possible while the calculator is switched off. Also, one sample is hardly enough to assert statistical evidence.

Hopefully calibrating the RTC against the HSI clock while running will be good enough. Using the HSE clock could be another, probably better solution.

sajjaadf:
Why won't this compile?

boricj:
My guess is that I did not add translations for languages other than French and English. You need to set the environment variable `EPSILON_I18N_LANGUAGES='en fr'` before building this branch.

sajjaadf:
oh okay

--
Sent with DejaLu (https://dejalu.me?sig)

boricj:
PoC pull request served its purpose.

PR #313, opened 2018-01-17
[ion] Re-calibrate the delay methods (msleep and usleep) 

EmilieNumworks:
The recent enabling of instruction and data cache increase the device performance which forces to recalibrate the delay methods

### DIFF
--- a/ion/src/device/device.cpp
+++ b/ion/src/device/device.cpp
@@ -20,12 +20,12 @@ extern "C" {
 // Public Ion methods
 
 void Ion::msleep(long ms) {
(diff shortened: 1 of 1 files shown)


boricj:
Wouldn't it be better to use the micro-controller's timers to keep track of time, at least for `Ion::msleep()`?

EmilieNumworks:
You are right, it would make msleep (and potentially usleep) independent from the compilation process (which is not the case right now). We could also count the cycle of the controller to measure the delay (SysTick). I'll add a comment for a future improvement!

Issue #196, opened 2017-10-08
Exam LED not blinking when not processing events


boricj:
When exam mode is activated, running the following Python program:

```
while True:
 pass
```

will hang the LED in its state as long as the program is running. In the same vein, powering down the calc while in exam mode will keep the LED unpowered.

PR #314, opened 2018-01-18
Ion::msleep and LED blinking with Timer + various register definitions

boricj:
Change `Ion::msleep` from a CPU busy-loop to a timer busy-loop. Ideally we'd put the MCU to sleep while sleeping to save on power, but we don't handle interrupts yet.

Also adding register definitions for #311 and a bugfix while I'm here.


boricj:
I cleaned up the pull request and switched `Ion::usleep` to use a timer too. Thus this obsoletes #313 with a cleaner solution that does not depend on how fast the CPU processes NOPs.

I won't attempt trying out interrupts without a means to debug the hardware though, so this pull request is done. Power consumption should be identical to the current implementation.

EmilieNumworks:
It looks like a very good PR, thank you @boricj ! It is not a good time right now for us to review and merge it as we are close to release a new update. However, if you could split the the first commit (about the Valgrind uninitialised read) from the others, I will merge it right away.

boricj:
Turns out setting up a timer for `Ion::usleep` takes long enough that the screen is significantly late for dimming, so I reverted that.

boricj:
While messing with timers, I've managed to decouple the exam LED blinking from the event loop and enable LED charging status. Activating exam mode makes the LED blink red and inhibits any LED change until it is deactivated.

A side-effect is that the LED will now keep blinking in exam mode while the calculator is shut down or busy (fixing #196). This might draw more power than before in this specific case, but the alternative would be far less pleasant for the student should an exam invigilator notices the lack of proper calculator blinkingness.

PR #316, opened 2018-01-22
Toolbox enhancement

EmilieNumworks:


poincare/src/random.cpp
```diff
@@ -0,0 +1,41 @@
...
+  assert(s == Sign::Positive);
+  return this;
+}
+
+template<typename T> T Random::random() {
+  if (sizeof(T) == sizeof(float)) {
+    uint32_t r = Ion::random();
+    return (float)r/(float)(0xFFFFFFFF);
+  } else {
+    uint64_t r;
```
EmilieNumworks: sizeof(T) == sizeof(double)

poincare/src/factor.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+Multiplication * Factor::addFactorsOfIntegerDecomposition(Integer i, Context & context, AngleUnit angleUnit) {
+  assert(!i.isZero());
+  i.setNegative(false);
+  Multiplication * m = new Multiplication();
+  if (Arithmetic::k_primorial32.isLowerThan(i) || i.isOne()) {
+    /* We do not want to break i in prime factor because it might be take too
+     * many factors... More than k_maxNumberOfPrimeFactors. */
+    m->addOperand(new Rational(i));
```
EmilieNumworks: Answer undefined in this case ?

PR #329, opened 2018-01-26
Full interrupt vector table definition, more registers and correct memory remap & vector table register after bootloader exit

zardam:
After leaving the bootloader, memory remap of address 0x00000000 is set to system flash memory, and vector table is defined at 0x1fff0000. After a reset, memory remap is set to main flash memory, and vector table is defined at 0x00000000. With this commit, reset values are forced at every boot (if not, interrupts are not working after bootloader exit).

Also, the vector table definition was expanded to the full list, and various registers definitions added.


Ecco:
CAUTION: this breaks our internal tools because the firmware header is relocated.

Ecco:
Forgot to say this is a very nice PR, thank you @zardam  I just commented in a rush about the header being relocated because this is something that we need to keep in mind 😃

PR #338, opened 2018-02-01
[emscripten] Fix save screenshot feature

adriweb:
It actually automatically downloads the screenshot now, as one would expect.


Ecco:
Looks great, thanks @adriweb :)

PR #339, opened 2018-02-01
Make cleanup

Ecco:
Clean the build system:
* Remove debug scripts from the root folder
* Put all configurable options in one place
* Use the epsilon name


notandxor:
The flash command has now been changed from: sudo make app_flash to sudo make epsilon_flash
The website still documents the former...

[https://www.numworks.com/resources/engineering/software/sdk/](url)

Ecco:
Indeed, we need to update the website.

PR #352, opened 2018-02-13
Use saner default debug flags

boricj:
Using `-O0` with the device target results in an ELF binary so bloated it won't fit in flash. Use saner `-Og` instead for debug builds.


Ecco:
This sounds like a great idea!

boricj:
So, any ETA for merging this?

Ecco:
Yes: now :)

Ecco:
The reason we didn't merge this sooner was because we wanted to make sure Clang had `-Og` too. Building on clang is broken right now, but that's something we'd like to restore in the short term. This will require the latest uPy though, since 1.9.3 isn't compatible with Clang just yet because of its use of naked functions.

PR #356, opened 2018-02-16
[apps] Shared: "Smart" incrementation of the first column of statisti…

claire-savinas:
…cs and regression apps

The default value of the first column is set to have the same increment as the two previous rows, when possible.
This allow faster data entry when the values of the first column are evenly separated.


EmilieNumworks:
This looks like a nice feature to add!

PR #359, opened 2018-02-16
Graph default range

EmilieNumworks:


poincare/src/expression.cpp
```diff
@@ -180,6 +180,15 @@ bool Expression::IsMatrix(const Expression * e, Context & context) {
   return e->type() == Type::Matrix || e->type() == Type::ConfidenceInterval || e->type() == Type::MatrixDimension || e->type() == Type::PredictionInterval || e->type() == Type::MatrixInverse || e->type() == Type::MatrixTranspose || (e->type() == Type::Symbol && static_cast<const Symbol *>(e)->isMatrixSymbol());
 }
 
+int Expression::polynomialDegree(char symbolName) const {
+  for (int i = 0; i < numberOfOperands(); i++) {
```
EmilieNumworks: By default, if numberOfOperands == 0, it should be -1.
Make a parent class number for Rational, Integer ...

Issue #197, opened 2017-10-09
sets deprecated?


yeKcim:
I try to compile epsilon git on manjaro.
```
sudo pacman -S git base-devel bison dfu-util freetype2 flex arm-none-eabi-gcc
git clone https://github.com/numworks/epsilon.git
cd epsilon
make clean && make
```
I got an error during make
```
poincare/src/expression_parser.y: avertissement: 7 conflits par réduction/réduction [-Wconflicts-rr]
QSTRDAT python/port/genhdr/qstrdefs.generated.h
I18N    apps/i18n.cpp
Traceback (most recent call last):
  File "apps/i18n.py", line 7, in <module>
    from sets import Set
ModuleNotFoundError: No module named 'sets'
make: *** [apps/Makefile:81: apps/i18n.cpp] Error 1
```

In apps/i18n.cpp line 7: `from sets import Set` can be replaced by `Set = set` but it seems that 

> In every recent python version sets are builtin as set and Python 3 got rid of the deprecated sets module altogether.

→ https://stackoverflow.com/a/12028231

AlainBusser:
:+1: 

The sets are a powerful and efficient way to introduce mathematical notions like

- the borelian tribe of events
- the probability of an event

I used them several years ago and it was[ surprinsingly easy](http://irem.univ-reunion.fr/spip.php?article526) :angel:

Also, to solve an equation consists in nothing else than compute the set of its solutions as exemplified by this function to solve the equations of degree 2:

    def solutions(a,b,c):
        Delta=b**2-4*a*c
        S = set([])
        if Delta>=0:
            r=Delta**0.5
            S.add((-b-r)/(2*a))
            S.add((-b+r)/(2*a))
       return S

Evariste Galois studied the permutations of this kind of sets and founded the group theory in this way...

adriweb:
@AlainBusser this actually has nothing to do with the epsilon-build-related python script issue @yeKcim is talking about :D
You should probably open a new issue and copy/paste your ideas (or paste it in https://github.com/numworks/epsilon/issues/162 directly?)

ian-abbott:
It should be easy enough to change the script to use the built-in `set()` constructor instead of `Set()`. However, the script has other things that fail when the `python` executable is Python 3. Perhaps the Makefile could run `python2` instead of `python`, but there might be systems that do not have the `python2` command.

AlainBusser:
[...]

PR #361, opened 2018-02-18
Fixes the build on hosts using Python 3

Ecco:
This update allows the build to succeed with both Python version 2 and 3


Ecco:
Fix #203 and fix #197

PR #362, opened 2018-02-22
[settings] Minimum of 5 m_cells in Settings::Subcontroller

### LINKED ISSUE #179: [bug] Assertion fail in Settings app with two languages
Building with `EPSILON_I18N_LANGUAGES='en fr'`, an assert is triggered on the simulator when accessing the `About` item in the `Settings` app:

```
app.elf: escher/src/table_view.cpp:125: int TableView::ContentView::typeIndexFromSubviewIndex(int, int) const: Assertion `typeIndex < m_dataSource->reusableCellCount(type)' failed.
```

On the device, the third item of the `About` menu is blank and the calculator crashes when exiting the app.

boricj:
Fixes #179


EmilieNumworks:
Thank you boricj, this code was indeed completely out of date and triggered bugs.

PR #369, opened 2018-03-21
Create a Storage that stores record (ex: python scripts)

EmilieNumworks:


apps/code/script_store.h
```diff
@@ -1,36 +1,36 @@
...
 class ScriptStore : public MicroPython::ScriptProvider {
 public:
-  enum class EditableZone {
-    None,
-    Name,
-    Content
-  };
-
   static constexpr char k_scriptExtension[] = ".py";
   static constexpr char k_defaultScriptName[] = "script.py";
```
boricj: Wouldn't a better name for an untitled script be `untitled.py`, like what most text editors on computers would do?

ion/include/ion/storage.h
```diff
@@ -0,0 +1,119 @@
...
+class Storage {
+public:
+  class Record {
+  friend class Storage;
+  public:
+    enum class ErrorStatus {
+      None = 0,
+      NameTaken = 1,
+      NonCompliantName = 2,
+      NoEnoughSpaceAvailable = 3,
```
boricj: Typo (should be `NotEnoughSpaceAvailable`).

ion/src/shared/storage.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+    if (s == r) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool Storage::nameCompliant(const char * name) const {
+  /* The name format is [a-z0-9_.]+ */
```
boricj: That regex does not match its implementation (only one dot is allowed, empty string allowed).

apps/code/script_store.h
EmilieNumworks: Good idea but we should localized 'untitled'. It is worth a new pull request!

PR #375, opened 2018-03-22
[ion] Fix updateModifiersFromEvent for emscripten

EmilieNumworks:
We want to call Termination event on emscripten platform without crashing.


ion/src/shared/events_modifier.cpp
```diff
@@ -24,7 +24,9 @@ void setShiftAlphaStatus(ShiftAlphaStatus s) {
 }
 
 void updateModifiersFromEvent(Event e) {
-  assert(e.isKeyboardEvent());
+  if (!e.isKeyboardEvent()) {
+    return;
+  }
```
EmilieNumworks: to do in emscripten specific files

PR #378, opened 2018-03-27
Config: add address sanitizer option for simulator in DEBUG=1

EmilieNumworks:


build/toolchain.host-gcc.mak
```diff
@@ -2,3 +2,8 @@ CC = gcc
 CXX = g++
 LD = g++
 GDB = gdb
+
+ifeq ($(DEBUG),1)
```
LeaNumworks, line 6: Make this optional?

LeaNumworks:
In another PR, I added ASAN by writing "export AFL_USE_ASAN = 1" in build/toolchain.afl.mak. Could you please check if this is also useful?

PR #380, opened 2018-03-30
[apps] Fix exam mode

zhaofengli:
05055c387f8b0535062bee4a0d5b93ddd86c8d2a seems to have broken the exam mode.

## Steps to reproduce
1. Open the native simulator
2. Try to enable Exam mode
3. The simulator crashes
```
epsilon.elf: apps/shared/interactive_curve_view_range.cpp:170: virtual void Shared::InteractiveCurveViewRange::setDefault(): Assertion `m_delegate' failed.
```

This is also present in the official Emscripten simulator at https://www.numworks.com/simulator/. On an actual device, the calculator simply reboots.

### DIFF
--- a/apps/sequence/graph/curve_view_range.cpp
+++ b/apps/sequence/graph/curve_view_range.cpp
@@ -75,11 +75,12 @@ void CurveViewRange::setTrigonometric() {
 }
 
 void CurveViewRange::setDefault() {
-  assert(m_delegate);
-  m_xMax = m_delegate->interestingXRange();
-  m_xMin = -k_displayLeftMarginRatio*m_xMax;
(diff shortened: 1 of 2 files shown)


adriweb:
I believe a more interesting thing to do is find out why `m_delegate` is NULL in the first place :P

zhaofengli:
@adriweb Sure. This only happens when the apps themselves (Graph/Functions and Sequence) are never opened since boot. `m_graphRange` never gets a delegate, which is set when the controllers are constructed.

Ecco:
Thank you very much @zhaofengli ! We had noticed that issue as well and came up with #382. Yours seems at least as good so we'll probably merge it :)

PR #383, opened 2018-04-01
[apps] Add numpad shortcuts to math_toolbox

zhaofengli:
This commit adds numpad shortcuts to the Math toolbox, as proposed in https://github.com/numworks/epsilon/pull/39#issuecomment-326604315. This makes it easier for experienced users to navigate the menu. Currently, items 1-10 are accessible through the numpad.

To prevent potential conflicts with the navigation logic in other menus, I've chosen to do this in the Math toolbox only for now. Some of the code is taken from the Python toolbox.

## GIF
numpad-shortcuts

### DIFF
--- a/apps/math_toolbox.cpp
+++ b/apps/math_toolbox.cpp
@@ -105,6 +105,23 @@ MathToolbox::MathToolbox() : Toolbox(nullptr, I18n::translate(rootModel()->label
 {
 }
 
+bool MathToolbox::handleEvent(Ion::Events::Event event) {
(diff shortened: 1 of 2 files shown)


EmilieNumworks:
Hi @zhaofengli, we prefer not to merge this. The shortcut enables the user to go to the 9th row but on the 10th, this behaviour is a bit odd and not standard. Feel free to keep a fork of the repository if you want to maintain this feature. Thanks anyway :)

PR #384, opened 2018-04-03
Make clena

Ecco:
Such an annoying typo…

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,4 +68,18 @@ clean:
 	@echo "CLEAN"
 	$(Q) rm -f $(products)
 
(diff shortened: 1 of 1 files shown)


boricj:
A bit late for an April's Fools. Why not leverage the `cowsay` program if it's installed (and perhaps print `Cows are mute on this system` if it's not available) ?

Ecco:
A bit late indeed :) That being said, this specific typo has been driving us mad for a while :-) We initially wanted to use cowsay, but dealing with the external dependency seemed like a lot of trouble for a practical joke :)

PR #409, opened 2018-04-20
Add ccache to Travis CI

chocol4te:
Takes the build time from ~6 minutes to ~4 at no additional cost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,12 @@
 language: cpp
+os: linux
+
+matrix:
+  include:
+    - env: PLATFORM=device EXT=elf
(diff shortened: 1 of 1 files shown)


chocol4te:
It would be great to setup Travis release deployment to GitHub, but that would require the repo owner to generate an OAuth key.

Ecco:
That's a very good idea, thank you for this @chocol4te !

Regarding the deployment to GitHub releases, what would be the use case? Manually flashing a firmware requires quite a bit of tooling already, so we thought that one might as well install the whole SDK anyway.

Ecco:
We moved away from Travis. Sorry we didn't merge this sooner 😐

PR #410, opened 2018-04-21
added better gamma for better color reproduction on LCD

tvillingett:
Better gamma curve, much improved color saturation on my device


tvillingett:
This is still not white balance adjusted, this could also be done in the digital gamma curves in the LCD driver chip, but I would need some assistance for testing it out, and some strategy how to color correct the screen.

tvillingett:
Seems like 0x00 gives the same gamma as 0x01 (G2.2) and it looks best of the ones, and is the default, so some more work with the fine tuning curves need to be done...

Issue #343, opened 2018-02-03
The toolbox pop-up's title is not localized


Ecco:
* Set language to French
* Press the "toolbox" key
* The title says "Toolbox".

![Screenshot](https://i.imgur.com/9TSC8IB.png)

UDXS:
This appears to be a design decision as it is written "Toolbox" in all language files but I have made a PR for it.

UDXS:
After testing it, this error has to do with the fact that the MathToolbox is not reinitialized after language change while PythonToolbox does it every time you open the python app so the translation applies. This also happens in the settings app once the language is changed where the window title does not change with it. I propose doing some form of system reset after language change, in the same way iOS does after one changes a language (black screen and text saying "Changing language...").

LeaNumworks:
Closing since this was solved by fd407ce3d9938072902d73de

PR #426, opened 2018-05-04
Fixed toolbox title localization

UDXS:
This is a fix for #343

I changed the translation files for the toolbox to say the name in the appropriate language. Unfortunately, this only worked in the Python toolbox. It had to do with the fact that the global MathToolbox would not get reinitialized when the language was changed.

**Fix Details:**
The toolbox i18n files have been updated. Now, every time the language is changed, a flag in GlobalPreferences is activated. When the MathToolbox is requested, it is initialized if it has to be (when a language changes or when it is created for the first time). It also frees up the memory used by the old toolbox.

### DIFF
--- a/apps/apps_container.cpp
+++ b/apps/apps_container.cpp
@@ -58,9 +58,9 @@ Poincare::Context * AppsContainer::globalContext() {
 }
 
 MathToolbox * AppsContainer::mathToolbox() {
-  if(GlobalPreferences::sharedGlobalPreferences()->languageChanged || m_mathToolbox == NULL){
+  if(GlobalPreferences::sharedGlobalPreferences()->languageChanged || m_mathToolbox == nullptr){
     GlobalPreferences::sharedGlobalPreferences()->languageChanged = false;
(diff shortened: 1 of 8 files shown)


UDXS:
Will fix it

UDXS:
I believe it is fixed now

UDXS:
Ok... Do not merge this yet. The problem has not been 100% solved.

UDXS:
The problem has to do with the fact that when the language is changed, the math toolbox does not get reinitialized, unlike to the python toolbox.

UDXS:
I have fixed the issue where the title text would not change with MathToolbox when the language was changed. It had to do with the fact that the global MathToolbox would not get reinitialized when the language was changed. 

**Details:**
 Every time the language is changed, a flag in GlobalPreferences is activated. When the MathToolbox is requested, it is initialized if it has to be (when a language changes or when it is created for the first time). It also frees up the memory of the old toolbox.

UDXS:
I'm closing this to fix the branches

PR #435, opened 2018-05-06
[apps] Code: Python syntax highlight

zardam:
This adds syntax highlight to the python code editor.

I had to disable the "dirty rect" optimisation in TextInput, in order to have the full text updated when typing a triple quote comment, but I somewhat enhaced the TextArea rendering, by clearing only what is needed, line by line. This way, the scrolling almost flicker free.

The color scheme and highlighting logic is based on what is done in [IDLE](https://docs.python.org/3/library/idle.html)

A sample of the result:

telechargement 2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@
 *.d
 
 # No lex / yacc generated files.
+escher/src/highlight_lexer.cpp
+escher/src/highlight_lexer.hpp
 poincare/src/expression_lexer.cpp
 poincare/src/expression_lexer.hpp
 poincare/src/expression_parser.cpp
--- a/apps/code/editor_controller.h
+++ b/apps/code/editor_controller.h
@@ -40,4 +40,3 @@ private:
 }
 
 #endif
-
--- a/apps/code/editor_view.cpp
+++ b/apps/code/editor_view.cpp
@@ -8,10 +8,16 @@
 EditorView::EditorView(Responder * parentResponder) :
   Responder(parentResponder),
   View(),
-  m_textArea(parentResponder, nullptr, 0, &pythonHighlighter),
+  m_highlighter(new PythonHighlighter()),
+  m_textArea(parentResponder, nullptr, 0, m_highlighter),
   m_gutterView(KDText::FontSize::Large)
 {
   m_textArea.setScrollViewDelegate(this);
+}
+
+EditorView::~EditorView() {
+  delete m_highlighter;
+  m_highlighter = nullptr;
 }
 
 void EditorView::scrollViewDidChangeOffset(ScrollViewDataSource * scrollViewDataSource) {
--- a/apps/code/editor_view.cpp
+++ b/apps/code/editor_view.cpp
@@ -1,13 +1,14 @@
 #include "editor_view.h"
 #include <poincare.h>
 #include <escher/app.h>
+#include <escher/python_highlighter.h>
 
 /* EditorView */
 
 EditorView::EditorView(Responder * parentResponder) :
   Responder(parentResponder),
   View(),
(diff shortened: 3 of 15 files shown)


adriweb:
Nice job indeed! Hopefully this gets merged :)

I also agree with the commit comment, it’s better if the class is generic (after all a “SyntaxHighlighter” is more fit for the standard text view that isn’t Python all the time, it’s just that sometimes, the highlighter would indeed be a PythonHighlighter).
For syntax coloring elsewhere, I guess mild colors for functions, but especially parens matching would be great!

zardam:
Thanks for the comments !

I added an "abstract" base class for the highlighter, and also parenthesis matching in another commit :

parenthesis

There is a little performance drop when moving the cursor. On the Mandelbrot script, when starting from the top of the script and keeping "right" pressed on two calculators, the cursor on the one with the parenthesis matching is 5 or 6 characters behind the other at the end of the script.

adriweb:
Nice update!

One comment from the compilers (I guess you just have to make it virtual):
```cpp
apps/code/editor_controller.cpp:25:3: warning: delete called on 'Highlighter' that is abstract but has non-virtual destructor [-Wdelete-non-virtual-dtor]
  delete m_highlighter;
  ^
```

Ecco:
This is, without a doubt, the best PR so far! Thank you @zardam 👍 That's obviously something we'll merge. We're on a very tight schedule these days, so please apologize for the delay and the succint feedback:

* Have you considered using uPY's own' parser? How about just its lexer then?
* I like the idea of leveraging Flex :)
* We'd really, *really* like to avoid malloc-ing a buffer. Would it be possible to do a single-pass lex and draw?
* We will most likely change the color theme

zardam:
> Have you considered using uPY's own' parser? How about just its lexer then?

I took a quick look at it, but I mainly focused on other "generic" highlighters (http://pygments.org/, https://highlightjs.org/, http://prismjs.com/, https://github.com/python/cpython/tree/master/Lib/idlelib). All of them are based on regular expressions. Using the parser can give a perfect result, but the output is not likely to be usable in case of an error in the syntax.

> I like the idea of leveraging Flex :)

I searched for some kind of static regular expression "compiler" but I found nearly nothing. So, flex seemed a good choice :)

> We'd really, really like to avoid malloc-ing a buffer. Would it be possible to do a single-pass lex and draw?

It seems possible to handle the buffer drawing inside the flex rules, this will avoid the "attributes buffer" completely, but some more work is needed (moving the logic of TextArea::ContentView::drawRect method inside flex rules). I will try to see if I can do it.

Ecco:
> Using the parser can give a perfect result, but the output is not likely to be usable in case of an error in the syntax.

Indeed. Using just the lexer seems like a good idea then. I think it'd be great to reuse uPy's to avoid code duplication.

zardam:
I checked the MicroPython lexer a bit more. It is certainly usable, but as it is designed to feed a parser, its output is not very convenient for the highlighting task : some tokens do not have an associated text and length, comments are ignored, some parts of the tokens are stripped, ... I will try to do something with it, but I think it will be far less straightforward than a few regexp.

zardam:
I have done something that is mostly working using the MicroPython lexer. There are still some minor glitches after the last character of the script, but that should be easily corrected.

The integration is not "clean" at all, as it is just a test for now.  

The result is here : https://github.com/zardam/epsilon/commits/test_upy_lexer

I can still spend some time to update this PR or cleanup the other implementation. Feedback/hints are welcome !

Ecco:
Hi @zardam ! I really liked your uPy PoC so I made a clean PR out of it. See #538.

PR #452, opened 2018-05-11
Saving Python script when home is pressed

claire-savinas:
otherwise, the modifications are lost.


EmilieNumworks:
Very good patch, thank you.

PR #453, opened 2018-05-12
Math Toolbox Title Localization Fix

UDXS:
This is a fix for #343

I changed the translation files for the toolbox to say the name in the appropriate language. Unfortunately, this only worked in the Python toolbox. It had to do with the fact that the global MathToolbox would not get reinitialized when the language was changed.

**Fix Details:**
The toolbox i18n files have been updated. Now, every time the language is changed, a flag in GlobalPreferences is activated. When the MathToolbox is requested, it is initialized if it has to be (when a language changes or when it is created for the first time). It also frees up the memory used by the old toolbox.

### DIFF
--- a/apps/apps_container.cpp
+++ b/apps/apps_container.cpp
@@ -58,9 +58,9 @@ Poincare::Context * AppsContainer::globalContext() {
 }
 
 MathToolbox * AppsContainer::mathToolbox() {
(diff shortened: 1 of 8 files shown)


LeaNumworks:
Hi, thanks for the PR and sorry for not reviewing it sooner.
I am closing it as it is outdated and #343 was fixed.

PR #478, opened 2018-05-16
[escher] Fix signed/unsigned int comparison

LeaNumworks:


escher/src/text_input_helpers.cpp
```diff
@@ -4,9 +4,12 @@
...
 
-int CursorIndexInCommand(const char * text) {
-  int textLength = strlen(text);
-  for (int i = 0; i < textLength - 1; i++) {
+size_t CursorIndexInCommand(const char * text) {
+  size_t textLength = strlen(text);
+  if (textLength == 0) {
+    return 0;
+  }
+  for (size_t i = 0; i < textLength - 1; i++) {
```
boricj: Might as well drop the strlen and iterate until we hit the NULL character. C++ short circuit evaluation will prevent buffer overflow.

escher/src/text_input_helpers.cpp
```diff
@@ -5,18 +5,15 @@
 namespace TextInputHelpers {
 
 size_t CursorIndexInCommand(const char * text) {
-  size_t textLength = strlen(text);
-  if (textLength == 0) {
-    return 0;
-  }
-  for (size_t i = 0; i < textLength - 1; i++) {
+  size_t index = 0;
```
adriweb: you probably meant `i` here,  and forgot to increment it in the loop?

LeaNumworks: Thanks! I'll just re-do the pull-request later :)

escher/src/text_input_helpers.cpp
adriweb: @boricj that made me wonder if the compiler optimized the strlen away by doing that internally (for simpler codes), in O2 or more, but I guess it can't if the implementation is specific or whatever :P https://godbolt.org/g/4RreRf

escher/src/text_input_helpers.cpp
adriweb: Note that you can force push an edited commit into the branch and the PR will be updated automatically, so I don't think you need to have another PR :)

escher/src/text_input_helpers.cpp
boricj: I recall an Underhanded C Contest entry where gcc optimized `strlen()` in a for loop down to O(n) but Microsoft's compiler didn't resulting in O(n^2), but I guess one must look at the disassembly to verify this.

PR #488, opened 2018-05-17
[escher] ExpressionLayoutFieldDelegate should not take a text as parameter but a layout

EmilieNumworks:
Warning: in ExpressionField: the serialized layout is not guaranteed to
be small enough to fill in the text buffer

### DIFF
--- a/escher/src/expression_layout_field.cpp
+++ b/escher/src/expression_layout_field.cpp
@@ -90,13 +90,13 @@ bool ExpressionLayoutField::privateHandleMoveEvent(Ion::Events::Event event, boo
     if (m_contentView.cursor()->pointedExpressionLayout()->removeGreySquaresFromAllMatrixAncestors()) {
       *shouldRecomputeLayout = true;
     }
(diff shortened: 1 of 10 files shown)


escher/src/expression_layout_field.cpp
```diff
@@ -120,10 +120,7 @@ bool ExpressionLayoutField::privateHandleEvent(Ion::Events::Event event) {
   }
   if (isEditing() && expressionLayoutFieldShouldFinishEditing(event)) {
     setEditing(false);
-    int bufferSize = TextField::maxBufferSize();
-    char buffer[bufferSize];
-    m_contentView.expressionView()->expressionLayout()->writeTextInBuffer(buffer, bufferSize);
-    if (m_delegate->expressionLayoutFieldDidFinishEditing(this, buffer, event)) {
+    if (m_delegate->expressionLayoutFieldDidFinishEditing(this, m_contentView.expressionView()->expressionLayout(), event)) {
```
EmilieNumworks: expressionLayout() instead of m_contentView.expressionView()...

PR #516, opened 2018-05-27
Update base.de.i18n

c-logic:
supports the charset german umlauts ?


vardai:
Should also be closed because of #647 and the issue currently being discussed in #673, @Ecco.

PR #522, opened 2018-05-29
[kandinsky] use pkg-config instead of freetype-config

### LINKED ISSUE #502: Use pkg-config instead of freetype-config in Makefile
Most modern linux systems have an up-to-date freetype version that does not provide this freetype-config exe anymore.

Instead, you should call `pkg-config freetype2`.

Tilka:
Since freetype2 version 2.9.1 freetype-config has been deprecated and is not installed by default anymore: https://sourceforge.net/projects/freetype/files/freetype2/2.9.1/

Fixes #502.


Tilka:
FYI, freetype2 has provided freetype2.pc since forever, so there should be no problem with compatibility.

Ecco:
Excellent, thank you @Tilka !

PR #534, opened 2018-06-03
HP Prime port

boricj:
... but I did it anyway.


boricj:
Note: the PRIME_OS.ROM file generated doesn't seem to work on real hardware. I can successfully sideload and run the ELF program with my GDB stub on the real hardware as well as running it inside @Gigi1237's emulator using a really old NAND image as a source. Will need to investigate this later.

Ecco:
😀 Next up, porting to Epsilon to your [toaster](https://www.embeddedarm.com/blog/netbsd-toaster-powered-by-the-ts-7200-arm9-sbc/)?

boricj:
> 😀 Next up, porting to Epsilon to your toaster?

Come on, that's *way* too obvious. Come to think of it, there's a couple of entry-level calculators from a couple of companies that have seen no software updates for a couple of years that could use an epsilon port and a couple of very, *very* snarky tweets...

A very tempting idea indeed, but I don't have that kind of free time to spend on overly elaborate practical jokes. Besides, I prefer my graphing calculators with *color* screens ^_^

debrouxl:
Entry-level calculators from two different companies which received no updates in the past few years... the HP 39gII and the Nspire CM (CAS), both of which use ARM processors ?

adriweb:
The HP 39gII doesn't have a color screen, though :P
And on the Nspire, an Nspire CX port would be more useful than a CM one, at least in terms of theamount of users... that said the port would likely be the same (except the LCD differences for W+ HW CX revisions)

boricj:
@debrouxl @adriweb You're way overthinking this. When I say I prefer my graphing calculators with *color* screens, that means I'm making fun of monochrome calculators, specifically the Casio Graph 25+E and TI-82 Advanced, both of which according to the latest TI-Planet's QCC haven't seen software updates since 2010 and 2015 respectively.

includereserved:
Thanks a lot for doing this. Does it support the touchscreen?

boricj:
It does not. We don't know how to use the touchscreen on bare-metal yet and escher has no concept/support for mouse/touch events anyways.

I intend to work again on this whenever I get a bit of free time and fix/clean stuff. Obviously we can't expect the NumWorks team to merge this and then support/maintain code to run on a _competitor's_ calculator, but I'd like to merge at least the platform-specific scaffolding (~first two commits) which are not HP Prime-specific.

I could unearth my RTC pull request and add a clock to this port too, but I don't want to go down the road of maintaining a full-blown fork either...

gerd2002:
Can we get a screenshot?

boricj:
There should be a couple of those on TI-Planet. I do intend to work again on this on the HP Prime G2 hardware someday as it'll hopefully be much less of a pain to work with for homebrews, but I need to get my hands on one first (and HP doesn't seem to be in a hurry to make those available in France).

boricj:
It's been fun and all, but I don't think I'll work on this specific port anytime soon.

RedGl0w:
> :grinning: Next up, porting to Epsilon to your [toaster](https://www.embeddedarm.com/blog/netbsd-toaster-powered-by-the-ts-7200-arm9-sbc/)?
> :grinning: Next up, porting to Epsilon to your [3DS](https://github.com/numworks/epsilon/pull/1528)

PR #537, opened 2018-06-04
Python syntax highlighting

Ecco:
Following a very interesting PR by @zardam #435 , here's an updated version.

* TextArea is not tied to Python anymore
* The Python heap is dynamically allocated (it risked overflowing the stack)
* If not enough memory is available, it falls back to black and white drawing
* Uses the MicroPython's lexer, thanks to a [very nice PoC](https://github.com/numworks/epsilon/compare/master...zardam:test_upy_lexer) by @zaradam
* Reuses and extends TextArea's dirty tracking
* Does syntax-highlighting on a per-line basis (avoids re-rendering the whole screen)
* The color theme that matches the one on https://workshop.numworks.com

Shortcomings:

* Doesn't do anything special for builtins because those aren't recognized by uPy's lexer.
* Doesn't treat function definition differently (because we're using a lexer and not a parser).

### DIFF
--- a/apps/code/Makefile
+++ b/apps/code/Makefile
@@ -11,6 +11,7 @@ app_objs += $(addprefix apps/code/,\
   editor_view.o\
   helpers.o\
   menu_controller.o\
+  python_highlighter.o\
   python_toolbox.o\
   sandbox_controller.o\
   script.o\
--- a/apps/code/editor_view.cpp
+++ b/apps/code/editor_view.cpp
@@ -1,4 +1,5 @@
 #include "editor_view.h"
+#include "python_highlighter.h"
 #include <poincare.h>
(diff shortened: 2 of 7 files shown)


adriweb:
The builtins seem pretty important, though.
Wouldn't something [like that](https://github.com/numworks/epsilon/compare/master...zardam:test_upy_lexer#diff-b8cb809527af879534ecc2fb0e0fe2e0R177) be good enough?

Ecco:
> Wouldn't something like that be good enough?

The problem here is that this code is somewhat reinventing the wheel. It'd be better to pull that info from uPy which already detects builtins.

Ecco:
Damn, wrong branch. Closing.

PR #538, opened 2018-06-04
Python syntax highlighting

Ecco:
Following a very interesting PR by @zardam #435 , here's an updated version.

* TextArea is not tied to Python anymore
* The Python heap is dynamically allocated (it risked overflowing the stack)
* If not enough memory is available, it falls back to black and white drawing
* Uses the MicroPython's lexer, thanks to a very nice PoC by @zaradam
* Reuses and extends TextArea's dirty tracking
* Does syntax-highlighting on a per-line basis (avoids re-rendering the whole screen)
* The color theme that matches the one on https://workshop.numworks.com

Shortcomings:

* Doesn't do anything special for builtins because those aren't recognized by uPy's lexer.
* Doesn't treat function definition differently (because we're using a lexer and not a parser).

### DIFF
--- a/escher/include/escher/solid_text_area.h
+++ b/escher/include/escher/solid_text_area.h
@@ -1,0 +1,30 @@
+#ifndef ESCHER_SOLID_TEXT_AREA_H
+#define ESCHER_SOLID_TEXT_AREA_H
+
+#include <escher/text_area.h>
+
+/* SimpleTextArea is a text area that draws text in a single solid color over a
+ * solid background. */
+
+class SolidTextArea : public TextArea {
+public:
+  SolidTextArea(Responder * parentResponder, KDText::FontSize fontSize = KDText::FontSize::Large,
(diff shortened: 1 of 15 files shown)


escher/include/escher/solid_text_area.h
```diff
@@ -0,0 +1,30 @@
+#ifndef ESCHER_SOLID_TEXT_AREA_H
+#define ESCHER_SOLID_TEXT_AREA_H
+
+#include <escher/text_area.h>
+
+/* SimpleTextArea is a text area that draws text in a single solid color over a
```
EmilieNumworks: Solid?

Ecco: See cd79d3731c3b4c2bb0c611b8b326e6c917f66849 😄

adriweb:
For those wondering what it looks like :)
screen shot 2018-06-04 at 21 59 21

PR #561, opened 2018-06-13
Fixed build on macOS

0b101:


0b101:
It appears that I am also having the same problems on my other Mac that is running Mojave. I have tried reinstalling freetype and the error doesn't disappear.

adriweb:
Haven't tried on Mojave yet, but I'll look at what freetype/pkg-config stuff I have installed, from brew...

In fact, we should probably have Travis launch mac builds as well, and see where it breaks.

.travis.yml
```diff
@@ -15,7 +15,10 @@ matrix:
     - env: PLATFORM=simulator EXT=elf TOOLCHAIN=host-clang
     - env: PLATFORM=blackbox EXT=bin QUIZ_USE_CONSOLE=1
 
-os: linux
+os:
+  - linux
+  - osx
```
adriweb: There's the whole brew setup to make in macOS' case, though.

adriweb: Also, the `os` part would have to go in the matrix conf.

adriweb:
No need to close it, we could have continued the investigation :p

adriweb:
So, on my mac I have (provided by brew):
`/usr/local/Cellar/pkg-config/0.29.2 (11 files, 627KB)` and `/usr/local/Cellar/freetype/2.9.1 (60 files, 2.6MB)`.
And the default latest Xcode 9.4 CLI tools installed.

Checking the flags:
```sh
> pkg-config freetype2 --cflags
-I/usr/local/opt/freetype/include/freetype2
> pkg-config freetype2 --libs
-L/usr/local/opt/freetype/lib -lfreetype
```

0b101:
Oh, I fixed it. All I had to do was `brew install pkg-config`.

PR #569, opened 2018-06-16
[apps] [i18n] Fix default language if not EN

adriweb:
When English was not built-in, this value wouldn't be available

### DIFF
--- a/apps/global_preferences.cpp
+++ b/apps/global_preferences.cpp
@@ -1,7 +1,7 @@
 #include "global_preferences.h"
 
 GlobalPreferences::GlobalPreferences() :
-  m_language(I18n::Language::EN),
+  m_language(I18n::Language::FirstAvailable),
   m_examMode(ExamMode::Desactivate),
   m_showUpdatePopUp(true),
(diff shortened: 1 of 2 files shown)


adriweb:
Note that I'm not sure the "Default" value is even needed anymore (just in case, I didn't touch it).
Also, the value "1" will always be defined anyway since using an empty list of languages in the build config will error beforehand

vardai:
You should also join issue #673 to help fix localization further.

adriweb:
Any update on that PR, by the way?
@Ecco @LeaNumworks @EmilieNumworks (not sure who to highlight :P)

LeaNumworks:
Hi @adriweb, I am sorry we did not review this PR sooner.
Thanks for spotting the problem and for your PR. However, I did another PR (#1454), which I think handles the problem a little better:
- As you suspected, the Default value was useless so I got rid of it
- Instead of adding another Language value, I modified the i18n.py script, so that the first locale passed as a parameter has the value 0, and the default language is now 0.

Thanks again!

Issue #584, opened 2018-06-26
[apps/graph] Hang when undef integral function


LeaNumworks:
![screenshot copie](https://user-images.githubusercontent.com/32390188/41897758-76bbb154-7928-11e8-8ad5-c4c3ecdf56c0.png)
In the Graph app, enter an integral function with an undefined symbol, such as int(a,1,2)
The device hangs when going in the Graph or the Table tab.

Issue found on [Reddit](https://www.reddit.com/r/numworks/comments/8tp8u2/crash_du_mode_fonction/)

LeaNumworks:
Fixed by #585

PR #585, opened 2018-06-26
Fix integral computation with undef

LeaNumworks:
Stop computing an integral if the evaluation of the function is undef.


LeaNumworks:
This fixes #584

PR #586, opened 2018-06-26
[apps/graph] Fix graph selection when a function is undef

LeaNumworks:
For instance, input f(x)=1/0 and g(x)=1, the graph tab now selects g
correctly


apps/shared/function_store.cpp
```diff
@@ -24,6 +24,22 @@ Function * FunctionStore::activeFunctionAtIndex(int i) {
   return nullptr;
```
Ecco: Use modelAtIndex(activeFunctionStoreIndex);

PR #595, opened 2018-07-01
adjust German wording in Settings app

mhaupt:
I propose changing EditionMode from "Eintrag" to "Eingabe", as the latter is more commonly used in German to describe the entry of text in a device. Also, "Linear" is a more apt description of "In Zeilen" (the latter suggests there are always multiple lines involved, which isn't strictly the case). Finally, "Natürlich" got rid of the trailing "e", which isn't necessary here.


Ecco:
Hi @mhaupt ! Thank you very much for your contribution! We have a hard time reviewing this because nobody speaks German on our team… Just like #647 .

Here's an idea: could you review issue #647, and I'll ask @vardai to review this one 😄 

Thank you so much again!

mhaupt:
That will be my pleasure @Ecco.

mhaupt:
Duly done. :-)

PR #605, opened 2018-07-08
Blending accuracy and speedup

ponceto:
Some code cleanup reordering and constification to help the C++ compiler.

In `KDColor::blend()`, the condition with alpha == 0xFF is tested first because this is the most common case, then alpha == 0x00. This speeds up the most common case.

In `KDColor::red()`, `KDColor::green()`, `KDColor::blue()`, the least significant bits are replicated to be more accurate when converted from 5/6 bits to 8-bits, so we have a quite good approximation from 5/6 bits to 8 bits component.

for example

`KDColor::red()` with a value of `0b1111100000000000` will return `0b11111111` and not `0b11111000`.
`KDColor::green()` with a value of `0b0000011110100000` will return `0b11110101` and not `0b11110100`.

and so on ... This is a well-known trick in computer graphics :-)

The code for blending has been improved to be faster (with less shifts and masks) an to give a more accurate blending (see the comments in `KDColor.cpp`)

### DIFF
--- a/kandinsky/include/kandinsky/color.h
+++ b/kandinsky/include/kandinsky/color.h
@@ -5,44 +5,76 @@
 
 class KDColor {
 public:
-  constexpr KDColor() : m_value(0) {}
-  // FIXME: This should not be needed, and is probably wasting CPU cycles
-  static constexpr KDColor RGB16(uint16_t rgb) {
-    return KDColor(rgb);
-  }
-  static constexpr KDColor RGB24(uint32_t rgb) {
-    return KDColor(((rgb&0xF80000)>>8)|((rgb&0x00FC00)>>5)|((rgb&0x0000F8)>>3));
-  }
-  static constexpr KDColor RGB888(uint8_t r, uint8_t g, uint8_t b) {
-    return KDColor((r>>3)<<11 | (g>>2) << 5 | (b>>3));
-  }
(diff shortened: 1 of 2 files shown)


adriweb:
You want to *rebase* the upstream commits, not merge them in your branch :)

ponceto:
@adriweb Sorry, made a mistake :-/

ponceto:
Correctly rebased ...

LeaNumworks:
Hi @ponceto, thanks for your PR and sorry for taking so much time to review it!
I tried it but, unfortunately, we will not merge it :

> In KDColor::blend(), the condition with alpha == 0xFF is tested first because this is the most common case, then alpha == 0x00. This speeds up the most common case.

Well, this would depend on the branch prediction technique. The best way to hint about likely branches would be to use `__builtin_expect`, which we haven't down until now.

> In KDColor::red(), KDColor::green(), KDColor::blue(), the least significant bits are replicated to be more accurate when converted from 5/6 bits to 8-bits, so we have a quite good approximation from 5/6 bits to 8 bits component.

This is interesting, but it adds operations and changes the behaviour, which we'd rather not do for now.

>The code for blending has been improved to be faster (with less shifts and masks) an to give a more accurate blending

I did a small benchmark and did not see increased speed with your changes.

Thanks anyway!

Issue #559, opened 2018-06-13
Incorrect syntaxic coloration in Python app


claire-savinas:
There are some problems with the syntaxic coloration in the Python app, check the end of the first lines :
![telechargement 9](https://user-images.githubusercontent.com/32409683/41333016-63ef3746-6ee0-11e8-9ade-0918db64ba5d.png)
![telechargement 8](https://user-images.githubusercontent.com/32409683/41333018-640b143e-6ee0-11e8-85d1-c9d8a4731396.png)
![telechargement 7](https://user-images.githubusercontent.com/32409683/41333019-6424d66c-6ee0-11e8-9991-8b9c72a8ecf1.png)

Ecco:
Indeed, good catch. Thanks @claire-savinas !

sagessylu:
On 11.2.0 version,
there are also problems with "\n" and "\t" when they are between quotes : 

![screenshot-2](https://user-images.githubusercontent.com/49091098/67139611-6c510580-f252-11e9-8563-dc9bf8f10234.png)

The second quote is grey when it's at the end of the line, else it is invisible.

LeaNumworks:
This has been fixed by 0f52335

PR #625, opened 2018-08-16
[apps/code] #559 Fix : Incorrect syntaxic coloration in Python app

claire-savinas:


Ecco:
Hi @claire-savinas ! Sorry for the (huge) delay commenting on this. That's quite an extensive review to make since it deeply modifies the highlighter's process. Namely, we're currently doing syntax-highlighting on a line-by-line basis. It probably doesn't work *all* the time, but it's very fast.

LeaNumworks:
Hi,
Sorry again for taking so much time to answer. I believe the syntactic coloration problems were fixed by 0f52335183a, while keeping the line-by-line coloration.
I am thus closing the PR, but many thanks for proposing this code.

PR #626, opened 2018-08-16
utime module

zardam:
This is a port of micropython utime module, without RTC related functions. Time keeping is done by using the systick counter of the CPU.


boricj:
If systick is an 1ms periodic interrupt, would it be possible to change `Ion::msleep` to a WFI loop? I assume this would save power while the calculator is idle but powered.

ion/src/device/boot/rt0.cpp
```diff
@@ -75,3 +75,9 @@ void start() {
 
   abort();
 }
+
+volatile long millis_elapsed = 0;
+
+void __attribute__((interrupt)) sysTick() {
```
Ecco: Are you sure this is required? GCC's doc says "On ARMv7-M the interrupt type is ignored, and the attribute means the function may be called with a word aligned stack pointer."

zardam: You're right, it is not required. There is a bit more information in the [ARM infocenter](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0774g/vvi1465810473505.html)

> In ARMv6-M, ARMv7-M, and ARMv8-M, the architectural exception handling mechanism preserves all processor registers, and a standard function return can cause an exception return. Therefore, specifying the interrupt attribute does not affect the behavior of the compiled output. However, ARM recommends using the interrupt attribute on exception handlers for clarity and easier software porting.

python/src/extmod/modutime.c
```diff
@@ -0,0 +1,123 @@
...
+
+#include "py/smallint.h"
+#include "py/runtime.h"
+
+long millis();
+long micros();
+
+void delay_ms(mp_uint_t delay) {
+    uint32_t start = millis();
+    while (millis() - start < delay && !micropython_port_should_interrupt()) {
```
boricj: Instead of busy waiting, do something like `Ion::msleep(std::min(millis()-start, 10))` (we don't have `std::min` implemented inside libaxx, but you get the idea)`

JFKingsley:
Any movement on this issue? Would be a really useful interface to have.

boricj:
Any ETA on this? I'm working on a native `turtle` module for Python which is coming along fairly nicely and I really, **really** need a clock to control the speed of the turtle in real time. This needs to be merged before I can submit a pull request for the module...

Ecco:
Yup, we definitely need to review this.

> If systick is an 1ms periodic interrupt, would it be possible to change Ion::msleep to a WFI loop? I assume this would save power while the calculator is idle but powered.

It might, but probably wouldn't be very significant because the backlight draws an order of magnitude more power than the CPU.

ion/include/ion.h
```diff
@@ -27,6 +27,8 @@ namespace Ion {
 
 void msleep(long ms);
 void usleep(long us);
+extern "C" long millis();
```
Ecco: Why the C linkage?

Ecco: Would it be possible to use an explicit type?
```suggestion
extern "C" uint32_t millis();
```

ion/src/simulator/init.cpp
```diff
@@ -116,3 +116,19 @@ void Ion::msleep(long ms) {
     }
   }
 }
+
```
Ecco: Couldn't we simply reuse the STL-based code from blackbox?

ion/src/emscripten/main.cpp
```diff
@@ -2,6 +2,9 @@
 #include "display.h"
 #include "events_keyboard.h"
 #include "../../../apps/global_preferences.h"
+extern "C" {
```
Ecco: Can't we re-use the STL code from blackbox?

ion/src/device/usb/calculator.cpp
```diff
@@ -9,6 +9,7 @@ namespace USB {
 namespace Device {
 
 void Calculator::PollAndReset(bool exitWithKeyboard) {
+  CM4.STCSR()->setTICKINT(false);
```
Ecco: Good catch. I think it'd make more sense to remove/reset the interrupt *before* calling `PollAndReset` though.

ion/src/device/regs/cm4.h
```diff
@@ -40,14 +40,39 @@ class CM4 {
     REGS_BOOL_FIELD(SLEEPDEEP, 2);
   };
 
+  class STCSR : public Register32 {
```
Ecco: I'd rename this to `SYST_CSR` as the ARM ARM calls it. I know, even ARM has two different names for this register, but the one with the most doc is `SYST_CSR`. Same goes for `SYST_RVR` and `SYST_CVR`.

ion/src/device/device.cpp
```diff
@@ -39,6 +39,22 @@ void Ion::usleep(long us) {
...
+}
+
+long Ion::micros() {
+    uint32_t c1 = CM4.STCVR()->getCURRENT();
+    uint32_t ms1 = millis_elapsed;
+    uint32_t c2 = CM4.STCVR()->getCURRENT();
+    uint32_t ms2 = millis_elapsed;
+    uint32_t load = CM4.STRVR()->getRELOAD();
+
+    return ((c1 > c2) ? ms1 : ms2) * 1000 + ((load - c2) * 1000) / (load + 1);
```
Ecco: Wow, that's some non-trivial math 😄 Maybe we should split that and add a few comments?

ion/src/device/device.cpp
```diff
@@ -184,9 +200,17 @@ void initPeripherals() {
 #endif
   Console::Device::init();
   SWD::Device::init();
+
+  CM4.STRVR()->setRELOAD(11999);
```
Ecco: I think this value deserves a quick comment. Why is that not 12000, by the way, since 96/8 = 12?

ion/src/device/device.cpp
```diff
@@ -184,9 +200,17 @@ void initPeripherals() {
...
+  CM4.STCVR()->setCURRENT(0);
```
Ecco: Maybe those should be wrapped in a `initSysTick()` method?

ion/src/simulator/init.cpp
zardam: In the simulator there are calls to 
  sDisplay->redraw();
  Fl::wait(0);
Without them, the content of the window is not updated in the while sleeping

ion/src/emscripten/main.cpp
zardam: No real reason to use SDL here. It was just included there

ion/src/device/usb/calculator.cpp
zardam: Ok

ion/src/device/device.cpp
zardam: Systick counter resets after counting to zero.12000-1 is needed to have exactly 12000 ticks between interrupts.

zardam:
I have updated the pull request from the comments. What I have not done for now is:
 - optimizing msleep (wfi or using systick) as I think it is a little bit tricky, there may be some cases where the function is called when systick interrupt is disabled. Maybe I could just check on entry and choose between implementations, but as said, the power consumption gain may not be worth it.
- factorizing millis() and micros() between non device implementations (may do it later, but it's too late for now)

Ecco:
Wow, amazing, thank you so much @zardam !

Now that I think about this, there is still a *major* issue with this PR, and it's a sneaky one.

The problem is with `delay_ms` and `delay_us`. In their current form, they pretty-much busy-wait. Sure, it's not very efficient, but that's not the main problem. The big problem is that this approach just won't work on Emscripten.

Imagine a small code that goes:

```python
while True:
  k = getKey()
  sleep_ms(100)
  move_with(k)
```

Handling blocking code in Emscripten has been a pain for a long time. The solution is to use Emterpreter. It does its magic and temporarily returns control to the JS main thread as long as you call `emscripten_sleep()`. But you *have* to call `emscripten_sleep`.

I think a simple solution could work out though:
* Have the `utime` code call `Ion::sleep` instead of doing its own thing
* Implement `Ion:sleep` with a call to `emscripten_sleep` on Emscripten, and with a call to SysTick on the device.

boricj:
@Ecco That solution will break keyboard interruption inside MicroPython while sleeping, we need to call `micropython_port_should_interrupt()` periodically to be able to break out. If we do end up indirectly calling `Ion::sleep()` in a loop from Python then we also need to adjust keyboard interruption handling as `micropython_port_should_interrupt()` will check the keyboard only every 20000 calls.

Ecco:
Good catch @boricj . Maybe a fix would be to call `emscripten_sleep(0)` at some point, which would yield a `setTimeout(.?.,0)`, giving the JS VM a chance to do other things in the meantime.

Ecco:
I'm wondering about the use cases for such a library. I'm concerned by how different it is from the CPython [time](https://docs.python.org/3.5/library/time.html#module-time) module.

How about starting up with a simpler version that actually only does `sleep(secs)` which is pretty much the only thing in common between `time` and `utime` anyway?

boricj:
Having access to a clock inside Python is useful for benchmarking and precise timekeeping inside animations and games. The API should consist of at least `sleep(secs)` and `clock()` (or a variant thereof) working both in floating point, to implement a useful subset of CPython's `time` module.

zardam:
I borrowed the module from MicroPython and the reason for the difference with CPython seems to be the lack of hardware floating point on embedded platforms (see at the bottom here : https://docs.micropython.org/en/latest/library/utime.html).

For the problem with sleeping in emscripten, maybe the solution is to add mixed calls to sleep, and micropython_port_should_interrupt (or a an immediate variant). I will try to do something in this way.

zardam:
I've done the sharing of millis & micros in blackbox and simulator. I also added a call to the sleep functions inside modutime.c, in order to still have the events processed while sleeping.

Emscripten build is broken. I'm trying to add calls to emscripten_sleep, but this must be done with Asyncify enabled, and this does not work for now.

zardam:
My bad, it was just needed to add the functions to EMSCRIPTEN_ASYNC_SYMBOLS in the Makefile. Emscripten build is working now, without blocking the main thread.

I had to rename at least usleep, because there was a collision with usleep from unistd (I also renamed msleep as well).

The whole thing is rebased onto current master.

zardam:
Last commit I expect :) I have changed the module to only include CPython functions (time, clock and sleep). I also reverted the usleep/msleep renaming as usleep is not used in C anymore.

ion/src/simulator/init.cpp
Ecco: Why would the window need to be updated *while* sleeping? There are no threads in Epsilon, so there's no reason for the window to change, is there?

zardam: It's to avoid this : 

image

This problem may be hidden by a compositor (on linux at least), which will keep the last state of the window.

This also allows to interrupt sleeping with the back key.

zardam:
"micros()" is not used for now, but I think it may be very useful for some kind of applications (games, ...)

adriweb:
Indeed - FYI even TI has a getMillis command in their lua, and it’s not like they’re aiming for game dev, it’s just useful overall ^^

Ecco:
@adriweb I removed `micros()`, not `millis()` 😄 

Several reasons:
* Microseconds (1/1000000th of a second) are very small, and I while I really see the use case for `millis()` I hardly see one for `micros()`. I doubt Python code can run so fast as to actually make a difference. And you always can busy-wait if needed, I doubt it's a big deal to block for such a small time.
* We're only using `Ion::usleep` on the Device, so this one will be removed from global Ion as well 😄

adriweb:
Woops, didn't noticed both were still there, nevermind my comment then :)

zardam:
It still may be useful in native code. If you want to run something at 60Hz (NES emulator for example :) ) you need to measure a period of 16.666 ms. Waiting 16 or 17 ms instead is pretty inaccurate.

Ecco:
Ok, so let's keep it in `Ion` then 😄  I'll remove it from the `time` module though.

ion/src/shared/timing.cpp
```diff
@@ -0,0 +1,9 @@
+#include <ion.h>
+#include <chrono>
+
+static auto start = std::chrono::high_resolution_clock::now();
+
+uint32_t Ion::millis() {
+  auto elapsed = std::chrono::high_resolution_clock::now() - start;
```
boricj: We should switch from `std::chrono::high_resolution_clock` to `std::chrono::steady_clock` to match the device's behavior.

Ecco: Good idea, thanks!

Ecco:
Oh, and I forgot the most important: thank you very much @zardam for this solid contribution!

PR #628, opened 2018-08-20
Fix warnings about comparing different signedness

Louson:
Some variables were declared as int while they are size_t. As we try to compare
them to unsigned values, a warning was raised (comparison of integer expressions
of different signedness).


Ecco:
Good catch! Thanks @Louson 😄

PR #634, opened 2018-08-24
Inverted SwitchView handle position

RemyDCF:
This pull request invert the handle position in SwitchView. It intends to make the switch a bit less confusing for new users, as it look more like switches on Android or iOS or other platforms.

Off state:
img_4562

On state:
img_2500


Ecco:
Excellent suggestion @RemyDCF, thank you!

PR #635, opened 2018-08-24
Revert "Changed active button color to YellowDark"

Ecco:
Reverts numworks/epsilon#602


Ecco:
Indeed, which is why this PR got droped :) See #636

PR #646, opened 2018-09-07
Corrected German and new Hungarian translation

vardai:
Hi,

I've corrected several important German translation errors and added a Hungarian translation.

Hope others may find my additions useful.

BR David


vardai:
Seems like German "ß" sends the compiler over the edge, although its use is significant. I don't even want to think about äÄ and Hungarian áéóöőúüűÁÉÍÓÖŐÚÜŰ. What should we do about it?

boricj:
I'd split your German corrections into a separate pull request. The NumWorks team has to this date not merged any user-contributed new language (nor introduced new languages since the initial release).

On a separate note, it reminds me that we need the ability to externalise translations in the same way we currently can put apps in separate, third-party repositories. Otherwise unofficial translations bitrot rather quickly...

Issue #598, opened 2018-07-02
[Minor] Serial number on calc is not in the same format as on the workshop


matheod:
On calc : letters + numbers

On workshop : only number

Would be better if worshop also used letters + numbers.

I didn't found what was the function to transform from one to the other.

Ecco:
Indeed. Serial number on the workshop are currently displayed as hexadecimal values. On the device, they are shown as Base64. It's the same underlying number though. We need to fix this on the workshop.

matheod:
Are you sure, I have only digit (no letter) on the workshop. And why I try to base64 decode, I get strange character ?

(the goal of this is to be able to find which of my student is the owner in case they ever mix up their calculator :D)

adriweb:
You probably want to decode the base64 to hex, not text/ascii :)

matheod:
> You probably want to decode the base64 to hex, not text/ascii :)

Oh you are right. And I was confused because my hex number only contain 0 - 9 digits :D

Ecco:
Solved 😄

PR #664, opened 2018-09-22
More friendly french translations

wow0000:
Edit of app/settings/base.fr.i18n with somes "de" to be a bit more friendly
`
SerialNumber = "Numero *de* serie" `
`
UpdatePopUp = "Rappel *de* mise a jour"
`
image

Thanks ! 👍


Ecco:
Very good suggestion. I think we didn't do this in the first place because we lacked some space (the serial number used to be longer, for example). We'll definitely need to check this. Thanks!

matheod:
Could be nice if the serial number on the workshop could also get the compression so they are both the same (or at least tell us how it's parsed). See #598

AlainBusser:
While you're at it please don't forget issue #215 :wink:

LeaNumworks:
Hi @wow0000,
Sorry for the huge delay to review your PR. I created #1462 with your rebased commit, because I could not push on your repo to change this PR directly, and I will merge that one.
Thanks for your code!

Issue #673, opened 2018-09-24
Accents and Diacritical Marks


vardai:
We really need to address accents and diacritical marks in Epsilon, the sooner, the better. Why? _Because it is the right thing to do,_ and there really is no alternative if we want this project to succeed in the long term. The important question how to do it properly, if the current structure of translations is adequate or how it has to be improved.

Please comment or react if you are in favor.

boricj:
While I would not say that the particular issue of accents and diacritics is a life-or-death issue for epsilon, the wider topic of i18n does warrant a discussion.

Epsilon's charset is currently 8-bit ASCII+proprietary set of ~24 symbols for codepoints above 0x7F. There's not enough flash for non-latin languages and all their glyphs, but even when reducing scope to western languages there are several things to consider before extending the charset:
- Keep symbols on proprietary 8 bits charset or move to "sparse" UTF-8,
- Use combined or combining characters,
- Add mechanism to input non-ASCII characters or keep them for UI only.

Personally I'd go for a sparse UTF-8 with combining characters for UI translations only and 7-bit ASCII+proprietary glyphs moved into ASCII control characters for everything else. It gives a unified system with the needed breathing room for translations while not forcing the entire system into variable byte-length encodings. This scheme should keep things under control so that compressing fonts would not be necessary in the short term to preserve Flash, but Python scripts from older versions might require a one-time conversion to the new scheme.

On the matter of translations, they are only in-tree. Third-party languages are condemned to bitrot in forks as shown time and time again, so a mechanism for out-of-tree translations like third-party apps is needed. Since out-of-tree translations are bound to lag behind, so a mechanism for optional sloppiness (missing translations being substituted by the first language) to not error out is probably needed too.

Ecco:
You definitely have a good point @vardai. Here the limiting factor is indeed the size in memory of rasterized glyphs, just like @boricj said, and he also raises good questions.

In the short term, translation files should use full-blown UTF-8, and the [conversion script](https://github.com/numworks/epsilon/blob/master/apps/i18n.py) should do the job of finding what is the best transliteration available given the current alphabet useable by Epsilon.

This allows to decorelate the technical limitations of Epsilon with the localizers' contributions. Localizer can say what the best output shall be, and then Epsilon does whatever it can to match the localizer's text.

For example, at the moment, since we only support ASCII characters, this script removes diacritics (`é` becomes `e`). I think a good first step would be to tweak this script to handle special characters for German.

[...]

PR #689, opened 2018-09-27
Hack for diacriticals

boricj:
As mentioned in #673, this is one way to quickly hack support for diacriticals in epsilon. This was quickly thrown together and is bound to overlook lots of edge cases, but it does show that it's possible to retrofit epsilon with accents at a reasonable cost.

image

This is obviously not to be merged as is. I do have one more comment though: **this is not the eighties anymore. Switch to Unicode and UTF-8 before somebody gets hurt.**


mhaupt:
Gotta love your additional comment, @boricj. ;-)

boricj:
PoC pull request hack served its purpose.

PR #692, opened 2018-09-28
[settings] Fix i18n

EmilieNumworks:


EmilieNumworks:
Hi @vardai or @mhaupt,
As you recently fixed a bunch of mistranslations to German, I refer to you. Do the following translations seem correct to you?
- Decimal = "Dezimal "
- Cartesian = "Algebraische "
- Polar = "Polar "

They now appear in 'Settings > Result Format' and 'Settings > Complex Format'.
Thank you very much!

vardai:
Hi @EmilieNumworks,
If the words are used in a choice kind of context, it would be 

- Decimal = Dezimal
- Cartesian = "Kartesisch" or "Exponentiell"
- Polar = "Polar"

You chose "Exponentielle" in your french translation. Let's then stick with your choice in German, too.

PR #695, opened 2018-10-01
[Sequences] Edit expression field is not erased in some cases

RemyDCF:
Hello,
I found a bug in Sequences, if you make a Un+1 sequence, the edit expression field is not erased between Un+1 and Un if you edit Un with any key except OK and Exe.

Here is an example:

This pull request intends to fix that.


LeaNumworks:
Hello @RemyDCF! Well spotted, thank you for fixing it!

PR #706, opened 2018-10-05
[Functions & Solver] Edit expression field is not erased in some cases

RemyDCF:
Hello,
I found a bug, and this is quite the same bug as in #695, but in functions and in solver.
The expression field is not erased after another field was edited.


EmilieNumworks:
Very good catch! However, when editing the expression of an expression list controller with another event than `OK` or `EXE`, we want to override the expression (instead of filling it) so we do not need to copy the previous expression before editing. The bug actually came from the fact that we forgot to clear the field of the input controller between two uses of if. #712 fixes this and reverts your previous fix on Sequence (#695).

RemyDCF:
Glad to see this is fixed. Thanks you very much!

Issue #716, opened 2018-10-09
matrix issue with the 10/09 1.7.0  release 


gnah33:
Minus signs just disappear when storing a matrix in memory.
try : [4 -1 2] sto M0
After that, M0 is [4 1 2].
Minus signs disappear whatever the matrix size.

Same problem on the 6/7th of Oct. version.
arm cpp compiler version is 7.3.1 (20180622)  if it helps.

LeaNumworks:
Hello! Good catch, thank you! This issue should be fixed by #718.

PR #718, opened 2018-10-10
[poincare] Fix Decimal::templatedApproximate

LeaNumworks:
Before, we used an unsigned mantissa and needed to set the sign
afterwards, but with the signed mantissa we must not do this.


LeaNumworks:
Fixes #716

PR #719, opened 2018-10-10
Font compression

Ecco:
Reduce the memory footprint of bitmap font by compressing them and reducing their dynamic range.

A comprehensive benchmark showed that LZ4 was the best compromise between:
* Light decompression routine
* Fast decompression
* Efficiency of the compression ratio

After this pull request, font drawing goes 2.5x faster.


Ecco:
For some reason, this creates a bug in the CurveView where the legend is not properly redrawn.

Ecco:
The CurveView bug is fixed by the last commit. What happened there was simple: we asked to draw only part of a glyph. Thanks to `KDContext::fillRectWithPixels`, which we should have used in the first place, the fix is trivial.

boricj:
I would assume icons/images could benefit from the same treatment. It might require rearranging the uncompressed image data layout to let LZ4 do a good job compressing these (maybe something like 4 bits per channel with each RGB channel stored as separate 4-bit grayscale bitmaps back-to-back).

The other obvious things to compress would be the preloaded python scripts (decompressed on reset), translations (the current language being kept decompressed in a RAM buffer) and `.data`, but the savings will probably be fairly modest in comparison.

PR #725, opened 2018-10-15
[i18n] german translation

EmilieNumworks:
- Wahrscheinlichkeit is too long to be displayed
- XPrediction describes the prediction of Y given X


EmilieNumworks:
@vardai and @mhaupt, thanks again for improving german translation. Two tiny issues with it:
- "Wahrscheinlichkeit" is too long to be displayed in the App menu. Can the short name "Warhsch." be acceptable?
- XPredicition is translated as "Prediction given X" in English (implied prediction of Y knowing X), so it should correspond to "Berechne Y", right?

mhaupt:
Abbreviating "Wahrscheinlichkeit" as "Wahrsch." is ugly but will have to do until the 4k display arrives on the Numworks device. ;-)

The "Berechne ..." solution makes sense. It's not 100 % accurate as it doesn't say from what the result is predicted, but I don't know a better way that's short.

PR #731, opened 2018-10-17
Add const modifier to primeFactors

boricj:
Saves 4000 bytes of RAM on non-LTO builds by leaving this array in Flash.


EmilieNumworks:
Good idea, thanks!

Issue #416, opened 2018-04-29
[Python] expand the APIs


UDXS:
Currently, NumWorks Python only has a few non-mathematical APIs. Those being kandinsky and some of the built in python ones. I suggest adding a few things. A good baseline minimum for where the API should be is what TI-BASIC offers. The glaring item that is missing is the inability to take keyboard input from the user. This is limiting especially if one wants to make a game or some form of interface. Speaking of interfaces and games, I think kandinsky's feature set should be expanded to allow easy shape drawing as well as text of different sizes (and maybe colors). On top  of this, a double buffering interface for smoother drawing would helo. As for interfaces, I think some form of simple interface module using escher that features simple buttons, check/radio buttons, input fields, and kandinsky drawing canvases would work well. Another helpful API would be Poincare. Being able to put an  expression string and have it solved or simplified would be great for computation. Finally, a way to store data would be helpful (Like a key-value store).

I don't believe that all of these features are required, or need to be done 100%. Either way, I think one can agree that the current Python APIs are a little sparse.

Thanks!

boricj:
There's also the Casio 90+E and its upcoming MicroPython port to consider. In a perfect world we'd share a common base API and we could have truly cross-platform calculator programs at last.

While I would not expect Casio to cooperate and coordinate with NumWorks on that subject, at least we'll be able to independently reimplement their base API once its out.

UDXS:
That's one way to do it. We can expand on it when it's done and (hopefully) end up with a better product. I'm currently working on Lua for microcontrollers and will port it to Epsilon. I hope to provide extensive APIs.

Ecco:
> Currently, NumWorks Python only has a few non-mathematical APIs. Those being kandinsky and some of the built in python ones.

That's a deliberate choice. Epsilon is a (quickly) moving target, and maintaining a stable API on top of this would be a significant burden on the development team.

> double buffering interface for smoother drawing would help

It would, and that is not restricted to Python. The problem is that the screen has quite a high resolution, so a single non-compressed frame is 320x240x2 = 153 KB.

> There's also the Casio 90+E and its upcoming MicroPython port to consider. In a perfect world we'd share a common base API and we could have truly cross-platform calculator programs at last.

On our side, we're completely open to starting a working group with other graphing calculators manufacturers to come up with a shared API.

[...]

PR #746, opened 2018-10-25
[python] Implement get_keys, wait_vblank in kandinsky

boricj:
These functions are already directly provided by kandinsky, might as well expose them to Python. Still a far cry from what we can do (#416), but these functions are essentially free to implement.


boricj:
I've added blitting and key reading functions, so one can make "real" games on the calculator now (proof: https://workshop.numworks.com/python/jblbeurope/pong).

Drawing performance is disappointing, probably because the framebuffer isn't memory-mapped, but it's usable as long as one doesn't attempt to redraw the entire screen in real time. The only way to achieve good drawing performance would be to "race" the beam by redrawing the entire screen in one go (like GESTE did), but that would be a quite the challenge to do it quickly enough from within MicroPython...

I've moved the Python interrupt key to the Home button. Not sure if that's a welcome change, but two main buttons is better than one and it's less likely someone would exit out of a Python script while trying to hit the OK key.

boricj:
I've added functions for blitting pixels (`get_pixels()`/`set_pixels()`) as well as reading keys (`get_keys()`). Performance is honestly quite terrible for several reasons (lack of memory-mapped buffer, display controller optimized for full-screen bulk updates) and racing the beam like in GESTE from MicroPython would be quite challenging to implement and use... On the upside, there's now everything we need to make real games in Python (proof: https://workshop.numworks.com/python/jblbeurope/pong).

I did change a couple of things, most notably the Python interrupt key for scripts (from Back to Home). On one hand accidentally hitting the Back key instead of the OK key while playing games is bound to infuriate some users and having two main action buttons is better than just one, on the other hand I'm not sure if that's really a good idea...

boricj:
I've updated the PR and the pong code sample to match #925. I also removed everything that isn't needed for that issue, except for key bindings changes in the simulator to make the Python app bearable with the keyboard.

I can't compile/test this on a real device for now, hopefully it'll work on the first try.

adriweb:
fill_rect is now available with 926b96cafbf7176bb39e78d7fa1617f25ddc254f + 8308068b694b397bbcc92ca3552541ec55ffa233

boricj:
@adriweb I noticed. I'm experimenting with tilemapping from Python right now in an attempt to justify a `set_pixels` method (and create games that look a bit better than those from an Atari 2600), but the non-intrusive ways I've tried are extremely non-optimal (MicroPython runs out of memory above 16 8x8 tiles, and that's just the tileset...). Overcoming this would require exposing epsilon's shared data storage as read-only files and adding text+binary record types, but I *highly* doubt the NumWorks team would be interested... Looks like the only thing we'll ever tilemap here will be the character table >_<

boricj:
I've updated the PR and removed what's implemented upstream now. The `get_keys` method returns a set of strings representing the currently pressed keys ; it's both easy to manipulate with set operators and fairly generic.

boricj:
Per #992, removing commit concerning simulator.

Ecco:
Hi @boricj ! Thanks for this PR!

I'm suggesting #1201 , see #925 for explanations.

elicolh:
be careful, up and right are inverted

Ecco:
Indeed, good catch. But this PR has been superseeded by #1201 anyway, so I'm closing this one.

PR #748, opened 2018-10-27
[python] Turtle module

### LINKED ISSUE #156: Add turtle graphics in Python
It can be of interest for teaching programming to students in a fun way and double as a showcase of the calculator's graphical and programming capabilities.

The reference is https://docs.python.org/3/library/turtle.html. The reference implementation uses Tkinter, so a port or a reimplementation on top of `kandinsky` would be required.

boricj:
This is a native implementation of the classic Python `turtle` module, complete with a cute lil' turtle icon. It's actually a fully native implementation in order to improve performance rendering. This currently adds a little bit over 2 KiB of Flash to the firmware.

```
from turtle import *
for i in range(255):
  gray=255-i
  color(gray,int(gray*0.75),int(gray*0.25))
  forward(i*0.1)
  left(10)
```
image

There are lots of improvements to do before this can be merged:
- [x] Add more turtle functions
- [x] Add alternative, shorter function names
- [x] Add toolbox menu for `turtle` module
- [x] Use a clock to control turtle speed in real-time (needs #626 merged first)
- [x] Allow interruption of turtle drawing (needs clock first)
- [x] Fix build dependency on turtle icon
- [x] (maybe?) Improve turtle logo rendering

Fixes #156.


Ecco:
Wow, definitely something we need to merge! Thank you very much for this awesome contribution @boricj ! We'll try to review this as soon as possible 😄

boricj:
Thanks @ecco!

Now that I look back on this, I do wonder if improving the Python runtime environment so that we can run the turtle module on top of it would be a better solution. After all, there's really nothing mandating a native implementation over a bytecode MicroPython module here, except for the lack of a decent Python platform API... Besides, calculator game programmers would very much prefer this option for obvious reasons.

boricj:
I've translated most of this native implementation to Python and it is indeed a much better solution. There's much less glue to maintain, the turtle implementation is looking far less scary in Python and the ability to interrupt a drawing comes "for free" since we're inside the MicroPython environment.

The downsize is that the sheer size of the `turtle` module written in Python won't fit in the 16 KiB MicroPython heap currently allocated. I need to figure out how to cross-compile the module to frozen MicroPython bytecode within the epsilon build system to make it fit.

Ecco:
> calculator game programmers would very much prefer this option for obvious reasons

Sorry, but they're really not obvious to me. Would you mind explaining why?

> I've translated most of this native implementation to Python and it is indeed a much better solution.

I'd love to see some benchmarks. Not only on speed, but also on memory usage (RAM + Flash). I'd assume the C++ version would win, but who knows 😄 Also, I don't find the implementation scary, and pretty much everything in Epsilon is C++ anyway…

python/port/modturtle_impl.cpp
```diff
@@ -0,0 +1,302 @@
...
+  return turtle_left(mp_obj_new_float(-mp_obj_get_float(angle)));
+}
+
+mp_obj_t turtle_left(mp_obj_t angle) {
+  float new_angle = ((t_heading - t_heading_offset) + (mp_obj_get_float(angle) * t_heading_scale)) / t_heading_scale;
+  return turtle_setheading(mp_obj_new_float(new_angle));
+}
+
+mp_obj_t turtle_goto(size_t n_args, const mp_obj_t *args) {
+  float newx, newy;
```
Ecco: Use mp_float_t which is a dynamic type. MicroPythons floats can either be single or double precision floats, and in our case they are actually doubles.
```suggestion
  mp_float_t newx, newy;
```

Ecco: Also, we might want to remove single-precision floats altogether at some point, so unless performance is critical I suggest using doubles when possible.

python/port/modturtle_impl.cpp
```diff
@@ -0,0 +1,302 @@
...
+  turtle_initpen(5);
+
+  return mp_const_none;
+}
+
+void turtle_initpen(int size) {
+  if (t_dot) {
+    delete[] t_dot;
+  }
+  t_dot = new uint8_t[size*size];
```
Ecco: We're on a `malloc` diet. Please avoid heap allocations.

kandinsky/src/context_rect.cpp
```diff
@@ -14,7 +14,7 @@ void KDContext::fillRect(KDRect rect, KDColor color) {
 }
 
 /* Note: we support the case where workingBuffer IS equal to pixels */
-void KDContext::fillRectWithPixels(KDRect rect, const KDColor * pixels, KDColor * workingBuffer) {
+void KDContext::fillRectWithPixels(KDRect rect, const KDColor * pixels, KDColor * workingBuffer, KDColor * prevPixels) {
```
Ecco: I'm rather uncomfortable modifying this, and I also have a hard time figuring out what you're trying to achieve here 😄 Would you mind explaining so we can see if there's a better solution?

boricj: In order to move the turtle across the screen, I need to erase the turtle icon by blitting back the pixels under it, so I need a way to fetch those pixels that's more efficient than brute-forcing `getPixel()`. This is a very dirty hack to write-back the pixels under the turtle, but I've changed it to a more conventional `fetchPixels(KDRect rect, KDColor * pixels)` in my current Python rewriting (not online yet).

boricj:
> > calculator game programmers would very much prefer this option for obvious reasons
> 
> Sorry, but they're really not obvious to me. Would you mind explaining why?

One upside to a turtle implementation in Python is that the platform API must be improved to accommodate it (more drawing primitives inside `kandinsky` and a stripped-down `time` module). As a side effect it will expose these improvements to user scripts, even if we don't officially acknowledge their existence in the toolbox. Besides supporting the turtle, these new functions will be useful mostly for calculator game programmers (the only missing piece would be a non-blocking `getKey()`, but that's another story).

> > I've translated most of this native implementation to Python and it is indeed a much better solution.
> 
> I'd love to see some benchmarks. Not only on speed, but also on memory usage (RAM + Flash). I'd assume the C++ version would win, but who knows 😄 Also, I don't find the implementation scary, and pretty much everything in Epsilon is C++ anyway…

Except for `speed(0)` (max speed), I do not expect speed to be an issue since the turtle will spend most of its time inside `time.sleep_ms()` and drawing will still be performed by native code. I'm not sure how memory usage will turn out, especially since there are a bunch of different technical solutions for storing the turtle icons with different trade-offs characteristics, but I doubt similarly optimized solutions will exhibit a large delta.

"Scary" is perhaps not the most adequate word here, but having the bulk of the code written in plain Python does result in a much prettier implementation overall and conveniently solves a bunch of issues for free. I'll make a pull request with that version as soon as possible.

Ecco:
> more drawing primitives inside kandinsky and a stripped-down time module

I see. Note that it's something we could do anyway 😄 

> the only missing piece would be a non-blocking getKey(), but that's another story

Indeed, but that would be a rather short story, since there's already a C++ API for this: `Ion::Keyboard::scan()`.

> I doubt similarly optimized solutions will exhibit a large delta.

There's one specific area I think the C++ version would be better: it could bypass Python-Heap allocation altogether, leaving more heap for end users.

kandinsky/src/context_rect.cpp
Ecco: I see 😄I think such an API could make sense in `KDContext` (a wrapper around `pullRect` and `absoluteFillRect`).

boricj:
> > the only missing piece would be a non-blocking getKey(), but that's another story
> 
> Indeed, but that would be a rather short story, since there's already a C++ API for this: `Ion::Keyboard::scan()`.

Unless I'm mistaken, `Ion::Keyboard::scan()` is currently a placeholder on the Emscripten platform since the keys won't work in the hardware test on the online simulator. It should be fairly easy to fix, but it's not quite a story as short as you think 😄

Ecco:
> Unless I'm mistaken, Ion::Keyboard::scan() is currently a placeholder on the Emscripten platform 

Indeed, you're a 100% right! We even have a [lengthy comment](https://github.com/numworks/epsilon/blob/7caf8681083cf096a26fd66b411c9ed4b22ec437/ion/src/emscripten/events_keyboard.cpp#L68) on the matter 😄

Ecco:
The more I'm thinking about this PR, the more I think I'd like the C++ version more than the Python one. It can be a great thing to actually compare both the C++ and the Python version, but all things equal I think I'd be more likely to eventually merge the C++ one: simpler build system, works like everything else, etc…

boricj:
I've opened a new pull request with the Python approach (#752). I think I'll close this pull request once feature parity has been reached as the other approach requires far less glue to work.

boricj:
> The more I'm thinking about this PR, the more I think I'd like the C++ version more than the Python one. It can be a great thing to actually compare both the C++ and the Python version, but all things equal I think I'd be more likely to eventually merge the C++ one: simpler build system, works like everything else, etc…

It boils down to a matter of personal taste, but I think the native approach requires way too much domain-specific glue. The Python approach does require a bit of work to make the MicroPython cross-compiler fit in the epsilon build system, but it's not _that_ bad and the side-effects are more useful (mechanism for additional bytecode modules in place for the future, generic improvements in kandinsky usable by everyone, keyboard interruption handled for free).

The native approach will have the edge on raw performance and memory consumption, but it's far more complicated to get right and it yields no improvements outside of the turtle module itself. Perhaps a hybrid approach (where drawing responsibilities are done by native code but the brain of the turtle is hosted by Python code) could bring the best of the two worlds...

boricj:
I could, but that's most useful when you have several turtles on screen, which this implementation will not support. The icon does have 8 different directions to represent headings, but the graphics are just placeholders for now subject to change.

LeSeulArtichaut:
Strangely I get an error while building (similar in both "v1" and v2)
v1 : `python/port/modturtle_impl.cpp:78:19: error: variable length array of non-POD element type 'KDColor'`
v2 : `python/port/mod_turtle_impl.cpp:46:24: error: variable length array of non-POD element type 'KDColor'`
Have I missed something ? This is possible !
Thought, about my gcc compiler :
```
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.4.0
Thread model: posix
```

boricj:
Sounds like your compiler handles PODs the old, pre-C++11 way. Ensure `-std=c++11` is set with a verbose build (`V=1`) and, if so, upgrade your toolchain 'cause it's buggy.

Ecco:
As you can guess based on the commits above, we're going to merge this in favor of #752 . Code is smaller, build system is simpler. Thank you very much @boricj for these great contributions!

Back on this specific PR, it's not ready to be merged as-is but we should be nearing completion. One of the difficulties is handling memory allocation properly. We decided to integrate with MicroPython's GC but it's not clear yet if we're really doing it the right way.

We also refactored the code so that there is a a Turtle class even though there will always be only one instance of it.

As a side note, we might end up temporarily removing entries from the catalog.

boricj:
No problem. Just bear in mind that the other version was more advanced and polished overall, especially on its handling of time. It might be worth back-porting things like `_stall` and the related algorithm of time/speed tracking because they are much more correct (and actually properly designed rather than hacked into submission).

python/port/mod/turtle/turtle.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include "turtle_icon.h"
+
+static constexpr KDSize k_iconSize = KDSize(9, 9);
+
+template <typename T> static inline T * allocate(size_t count) {
+  /* We forward dynamic allocations to the Python GC so we don't have to bother
+   * with deallocation. For this to work well, the Turtle object who owns the
+   * allocated areas need to be added to MicroPython's GC roots, otherwise those
+   * buffers will be wiped out by the GC too early. */
+  return static_cast<T*>(m_malloc(sizeof(T) * count));
```
boricj: What happens if MicroPython run out of heap?

Ecco: Well, literally nothing! Dynamic memory is used for drawing operations. If a given allocation fails, the matching operations are skipped!

python/port/mod/turtle/turtle.cpp
```diff
@@ -145,12 +139,38 @@ void Turtle::setPenSize(KDCoordinate penSize) {
...
   if (m_visible) {
-    draw();
+    draw(true);
   } else {
     erase();
   }
 }
 
+void Turtle::setColor(const char * color) {
+  if (strcmp("blue", color) == 0) {
```
Ecco: How about something along the lines of
```
class Pair {
public:
  constexpr Pair(KDColor value, const char * name);
};

Pair pairs[] = {Pair(KDColorRed, "red"), Pair(KDColorBlue, "blue"),...};
for (Pair p : pairs) {
  if (strcmp(p.name, input)) {
    m_color = p.color;
  }
}
```

python/port/mod/turtle/turtle.h
```diff
@@ -76,12 +79,26 @@ class Turtle {
...
   static constexpr uint8_t k_defaultPenSize = 1;
 
+  enum class PawType : uint8_t {
+    FrontRight = 0,
+    BackRight = 1,
+    FrontLeft = 2,
+    BackLeft = 3
+  };
+
+  enum class PawPosition : int {
```
Ecco: Why bother specifying `int` as a backing type?

LeaNumworks: I forgot to clean this ! Done with 91c4fea

python/port/mod/turtle/turtle.cpp
LeaNumworks: Done with 99d71bb!

escher/include/escher/toolbox_message_tree.h
```diff
@@ -5,20 +5,40 @@
...
+  constexpr static ToolboxMessageTree Leaf(I18n::Message label, I18n::Message text = (I18n::Message)0, bool stripInsertedText = true, I18n::Message insertedText = (I18n::Message)0) {
+    return ToolboxMessageTree(
+        label,
+        text,
+        (insertedText == (I18n::Message)0) ? label : insertedText,
+        nullptr,
+        0,
+        stripInsertedText);
   };
+  constexpr static ToolboxMessageTree Node(I18n::Message label, const ToolboxMessageTree * children, int numberOfChildren) {
```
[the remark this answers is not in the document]
LeaNumworks: Indeed, thanks! Done by df2b9c4.

gpotter2:
Very cool @boricj ! Looking forward to the release

PR #752, opened 2018-11-06
[python] Turtle module, version 2

boricj:
See #748 for context. I think I'll close the original pull request when this one reaches feature-parity as this approach is far cleaner overall.


boricj:
I've fleshed out this pull request as far as possible so that it has at least a fighting chance against #748. Note that it is essentially complete unlike the other version: turtle speed is correctly metered in real-time, keyboard interruption works properly and so on.

I've settled on a hybrid approach that keeps drawing on the native code side in order to keep the Python brain uncluttered. As stated before, main advantages are:
- keyboard interruption while turtling Just Works(TM) as long as the `time` module does its job
- minimal amount of native glue
- uncluttered turtle implementation in Python without having to worry about native issues
- (debatable) probably less costly to maintain in the long run overall

This does consume some Flash, currently at 10959 bytes or 10.7 KiB when compared against master whereas the native implementation clocks in at 6525 bytes or 6.4 KiB. It is more than I've expected. Accounting for the fact that the native version is less mature/featureful and is bound to balloon a bit + there are ways to trim the Python down if necessary (we can reclaim probably about 0,25 KiB just with a tighter naming of variables/imports alone), it's not a _huge_ gap.

Also, I can't quite seem to be able to nail the CI for some reason... It does build and work without issue on the simulator and on the device on my side.

boricj:
I've rebased everything according to the latest changes and removed everything that doesn't have to be there. Firmware flash size difference between master and this pull request is down to 7098 bytes, or a tad under 7 KiB.

Ecco:
Hi again @boricj ! We really appreciate both contributions, but we had to pick one anyway. We know you liked the Python one better, but it's heavier and introduces some more complexity in the build process, so we opted for #748 . Thank you very much!

PR #754, opened 2018-11-08
Emscripten keyboard scan

Ecco:
Implement `Ion::Keyboard::scan` on Emscripten


boricj:
Small nitpick: I don't think this will interrupt a Python script when the user presses the ESC key on its (physical) keyboard.

Ecco:
Indeed, it doesn't. It also doesn't expose the computer's physical keyboard keys (if any) to `Ion::Keyboard::scan` either. But it's a start 😄

PR #755, opened 2018-11-08
Emscripten python live drawing

Ecco:
Allow long-running Python scripts to refresh the display

### DIFF
--- a/python/port/helpers.cpp
+++ b/python/port/helpers.cpp
@@ -12,11 +12,11 @@ void micropython_port_vm_hook_loop() {
   /* Doing too many things here slows down Python execution quite a lot. So we
    * only do things once in a while and return as soon as possible otherwise. */
   static int c = 0;
-  c++;
-  if (c % 20000 != 0) {
+
+  c = (c + 1) % 20000;
+  if (c != 0) {
     return;
   }
-  c = 0;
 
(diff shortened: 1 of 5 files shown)


python/port/helpers.cpp
```diff
@@ -4,15 +4,23 @@ extern "C" {
...
+  /* This function is called very frequently by the MicroPython engine. We grab
+   * this opportunity to interrupt execution and/or refresh the display on
+   * platforms that need it. */
+
+  /* Doing too many things here slows down Python execution quite a lot. So we
+   * only do things once in a while and return as soon as possible otherwise. */
   static int c = 0;
   c++;
-  if (c%20000 != 0) {
+  if (c % 20000 != 0) {
```
boricj: Might as well replace this with something a bit cleaner:
```c
static int c;

c = (c + 1) % 20000;
if (c != 0) {
  return;
}
```

Ecco: I'm sorry, but dumb me doesn't understand what's different in what you suggest 😄

boricj: The behavior is (nearly) identical, it's just more concise and a bit easier to read. It's probably just a matter of personal taste though.

LeaNumworks: Thanks your for the suggestion! I made a commit cleaning this piece of code.

Issue #609, opened 2018-07-12
washed out colors / gamma correction / color correction


tvillingett:
Hi, as several others have commented the display seems to be a bit washed out. I think it could be a bit better with some gamma correction, and I think to do it directly in the LCD driver chip would be the best to avoid losing colors. I have looked through the data sheet of the controller, and it seems quite a lot to setup, and maybe to use code from other linux could help:
https://github.com/torvalds/linux/blob/63f04777162181798399a2c4e5436d0d0c16291b/drivers/staging/fbtft/fb_st7789v.c#L172

boricj:
Reusing GPL-licensed is not an option here. Reimplementing is the way to go (and it really doesn't look bad at all, it's just populating a set of registers with values).

tvillingett:
yes, sure that could be done, the other isssue is how to interpret the gamma curve from the numbers. I cannot see the logic in the numbers, it almost follow a line/curve except for 48, 33, 53 which looks like the top of a cowboyhat...
[gammacurve_original.pdf](https://github.com/numworks/epsilon/files/2190258/gammacurve_original.pdf)

claire-savinas:
To understand the values, you have to look in the [datasheet](https://www.numworks.com/shared/binary/datasheets/st7789v-lcd-controller-73f8bc3e.pdf) of the display controller, page 287. Some values are "packed" (8th parameter, for example) and not in the right order.

Once done, you have to look at page 139 to have the relation between the values and the voltage applied to the LCD.

I'm asking myself if the datasheet is correct. According to it, VP7 to VP12 are equals (and also VP51 to VP56). This seems strange...

I have put together a [cell sheet](https://github.com/numworks/epsilon/files/2316479/gamma_numworks.zip) with the interpreted formulas, if you are interested, and also tested some values, but I don't know if the result is really better than the default ones. If you want to try : fe3a7bf

tvillingett:
Sorry for my late response, but I finally got a chance to compile and test your suggested change, and for me it really made a difference. The orange is much more saturated for example, more like the simulator on a computer screen, and the light blue in the background in the calculator screen really shows. It is hard to make a photo, but I will try. So the only adjustments I would like to try, is to reduce the blue a bit to get more neutral white/grey and perhaps put back the "gamma" correction you made a notch... Thanks a lot! I will look into making the adjustments.

tvillingett:
[...]

PR #756, opened 2018-11-10
[ion/device/display] Gamma correction (#609)

claire-savinas:
An attempt to have a better color rendering on the device.

Gamma correction value were calculated by @tvillingett


Ecco:
Hi! I just gave this a try and this looks a lot *worse* 😄
I tried to take a photo, the result isn't that good, but you get the idea. Left is before, right is with the patch applied. The colors are completely desaturated and washed-out…

img_0759

tvillingett:
Hmm.
Do we have different displays/ or batches perhaps? I could not find the display manufacturer in the data sheets for the Numworks hardware.

I tuned the values a slight bit with the values below:
```c++
  #define ARR_GAMMA_AE_NEW3 0x70,0x5,0x9,0x9,0xB,0x1A,0x30,0x66,0x3D,0x15,0x10,0x10,0x18,0x1F
  SEND_COMMAND(PositiveVoltageGammaControl, ARR_GAMMA_AE_NEW3);
  SEND_COMMAND(NegativeVoltageGammaControl, ARR_GAMMA_AE_NEW3);
```
And the look for me is almost identical with the one you have to the left.
Can we make a dump from the memory of the chip to make sure the default value is what we think it is? What else could be different?

tvillingett:
One difference: I have made my changes to an old branch, with the Swedish translation, but it does not seem to be different in the display init section.

Ecco:
Hi! Thanks for this PR! Unfortunately I'm going to close it because we've already merged something similar with [per-panel gamma calibration](https://github.com/numworks/epsilon/blob/587b5e7a930593c42f7fdbd16bce19b9fe04380e/ion/src/device/shared/drivers/display.cpp#L434).

PR #773, opened 2018-11-17
Cast result of strlen to int to match comparison

Aposhian:
There are three locations in code that generate a compiler warning because of a comparison of unsigned integer with signed integer types. This is because `strlen` returns `size_t`. I cast these instances to `int`. This silences the compiler warnings.


LeaNumworks:
Hi @Aposhian, I am very sorry we took so much time to review your PR.
Thanks for your commit, however the code is outdated now so I am closing the PR.

PR #780, opened 2018-11-23
[build] Disable WebAssembly on emscripten

LeaNumworks:


adriweb:
Remove the disabling you mean? Unless I'm not understanding that arg :P

boricj:
What's the reason for disabling WebAssembly here?

Ecco:
It's not a final decision, but at the moment there are still [~20% users](https://caniuse.com/wasm) who cannot use WebAssembly, so we would still need a pure-JS version to avoid regressions.

Now in our case WebAssembly brings little benefits:
* IIRC the resulting "binary" isn't much smaller than the JS version
* It may be faster, but even the JS version is an order of magnitude faster than the device
* It would make the offline version bigger since it should embed both the JS and the WASM version

Long story short, WASM looks cool but doesn't justify the extra work at the moment.

PR #798, opened 2018-11-29
Poincare reduce

EmilieNumworks:
Add a flag to the Poincare reduce routine to choose between a strict and a relaxed reduction.
For instance: x^0 --> 1 only in relaxed reduction.

### DIFF
--- a/poincare/src/power.cpp
+++ b/poincare/src/power.cpp
@@ -140,8 +140,8 @@ int PowerNode::serialize(char * buffer, int bufferSize, Preferences::PrintFloatM
 
 // Simplify
 
-Expression PowerNode::shallowReduce(Context & context, Preferences::AngleUnit angleUnit) {
-  return Power(this).shallowReduce(context, angleUnit);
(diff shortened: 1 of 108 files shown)


poincare/src/power.cpp
```diff
@@ -331,12 +331,14 @@ Expression Power::shallowReduce(Context & context, Preferences::AngleUnit angleU
         replaceWithInPlace(result);
         return result;
       }
-      /* Warning: in all other cases but 0^0, we replace x^0 by one. This is
-       * almost always true except when x = 0. However, not substituting x^0 by
-       * one would prevent from simplifying many expressions like x/x->1. */
-      Expression result = Rational(1);
-      replaceWithInPlace(result);
-      return result;
+      if (!strict) {
```
LeaNumworks: Even if strict == true, you might want to get rid of 3^0 or log(4)^0. 
You could for instance:
- Get rid of the power if the evaluation of childAtIndex(0) is not near 0
- Get rid of the power if childAtIndex(0) is a Number not 0, and other simple cases

PR #799, opened 2018-11-29
[shared] Fix the buffer size of curve view labels

EmilieNumworks:


apps/shared/curve_view.h
```diff
@@ -41,6 +41,7 @@ class CurveView : public View {
   constexpr static KDCoordinate k_okVerticalMargin = 23;
   constexpr static KDCoordinate k_okHorizontalMargin = 10;
   constexpr static KDCoordinate k_labelGraduationLength = 6;
+  constexpr static int k_labelBufferSize = 7;
```
LeaNumworks: You could compute the value :)

PR #844, opened 2019-01-01
Add display color inversion

DSchndr:
Since there is no dark theme for using the calculator "in the dark" i thought that this feature is quite useful.

It adds color inversion (which is available in the display controller) to ION and also an switch in the settings app to enable/disable color inversion.

Only thing that isn't done is translations, because i can't speak every language :)


sajjaadf:
GREAT

sajjaadf:
can we merge this now?

sajjaadf:
does this work in simulator?

DSchndr:
You can try to merge and check it out or directly use inverted branch from me.
It should build without problems.
It *only* works on the calculator and *not* in the simulator because it is using an command from the lcd controller (simulator doesn't simulate this unfortunatly)

Oh and if you don't mind @sajjaadf, could you please use the edit option instead of writing multiple comments.

sajjaadf:
sorry

IonImpulse:
This would be great to have, especially for late night studying. Thanks!

Ecco:
Hi @DSchndr ! Thank you very much for this contribution! Unfortunately we don't have a plan to support different themes / look'n'feels at the moment, so I'll have to close this pull request. Thank you very much for your time though…

PR #863, opened 2019-01-16
[python/turtle] Store the heading in degrees with no k_headingOffset

LeaNumworks:
This reduces rouding errors, such as:
setHeading(0)
right(1)
heading() -> gives -1.00...01

### DIFF
--- a/python/port/mod/turtle/turtle.cpp
+++ b/python/port/mod/turtle/turtle.cpp
@@ -31,7 +31,7 @@ void Turtle::reset() {
   // Reset turtle values
   m_x = 0;
   m_y = 0;
-  m_heading = k_headingOffset;
+  m_heading = 0;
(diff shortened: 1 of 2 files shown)


python/port/mod/turtle/turtle.cpp
```diff
@@ -45,20 +45,18 @@ void Turtle::reset() {
 
 bool Turtle::forward(mp_float_t length) {
   return goTo(
-    m_x + length * sin(m_heading),
-    m_y + length * cos(m_heading)
+    m_x + length * cos(m_heading * k_headingScale),
```
EmilieNumworks: is it std::cos?

LeaNumworks: Fixed!

PR #898, opened 2019-02-12
Fix for German typos in variables.de.i18n

RomanWu:
"Expressions" was not German so far, I suggest "Ausdruecke"
The correct plural of function in German is "Funktionen".
"Namen" is a noun, and therefore written uppercase in German.


EmilieNumworks:
Good catch, thank you Roman!

PR #904, opened 2019-02-19
Poincare clean

EmilieNumworks:


adriweb:
The commit message of 932d647 is wrong, it seems - I think you mean `Opposite::Builder(Opposite::Builder(Rational(1))) creates "--1"`?

EmilieNumworks:
Oups, you're right. I'm fixing it! Thanks

Issue #902, opened 2019-02-15
Increase the number of brightness levels. 


varszegimarcell:
Hi 

I have a suggestion, for brightness customization. 

When you open the settings app,  you can choose five brightness levels. In my experience, this isn't precise enough, as some settings in the middle (2/5 or 3/5) can become too bright in darker environments, as well as too dark in lamplight etc. 

This problem can be easily solved, by increasing the values of brightness levels, allowing more precise backlighting settings. 

You just have to change one constant in the apps/global_preferences.h header file. 

`23: constexpr static int NumberOfBrightnessStates = 5; // original code`

My suggestion is, to change this value to 20.

`23: constexpr static int NumberOfBrightnessStates = 20; // my recommendation `

I've tested this "mod" on my calculator, works perfectly fine. 

Thanks for your amazing work on this great piece of tech, this calculator made my math studies much easier. 

Best regards.
Marcell Varszegi

INailedIt:
Have you tried committing your mod?

INailedIt:
I phrased that horribly.

varszegimarcell:
Can I commit to this project? Do I have the rights for this?

adriweb:
You can commit to your fork of this project. From which you can make a PR, that is, a suggested change to be merged here :)

Ecco:
Note that the hardware can only do 16 levels of brightness. Anything higher than that won't have any effect.

varszegimarcell:
> Note that the hardware can only do 16 levels of brightness. Anything higher than that won't have any effect.

Oh I see. I missed that some of the brightness levels doesn't change anything if the constant is setted to 20.

> You can commit to your fork of this project. From which you can make a PR, that is, a suggested change to be merged here :)

Thanks, now I know how the git system works. I'm studying computer science engineering, and we learned about software development teamwork in this semester. Hopefully, I will became a good software engineer one day... :D

As this feature already added to the project, I close this issue thread.

PR #910, opened 2019-03-10
Brightness settings

jperon:
dimBacklightBrightness, idleBeforeDimmingSecs and idleBeforeSuspendMins parameters
with entries in settings menu.
This allows the user to change time before dimming screen or suspending, and to set the level of dimmed brightness.


jperon:
N. B. : incidentally, fix #902.

Ecco:
Here are a couple screenshots of what this PR adds.

**Main settings menu**
Main menu

**Brightness menu**
Brightness menu

Ecco:
I personally think this is an interesting PR, but I'm not sure wether we should merge it. It adds quite a lot of complexity, and I'm not quite sure which problem this is trying to solve.

As a general rule, we try to add as _few_ settings as possible. Is there a reason one would like to change the delays and the dimmed brightness?

adriweb:
> As a general rule, we try to add as _few_ settings as possible. Is there a reason one would like to change the delays and the dimmed brightness?

I suppose this would make it aligned with other calcs (at least, high-end ones) that provide this kind of tweaking in their settings.
For instance on the TI-Nspire CX:
image

(The brightness itself it done via two keypresses)

boricj:
I think the "Dim" setting is superfluous, as simply subtracting a couple of brightness levels from its normal level when dimming should be enough. Similarly, the "Dim after (s)" setting is probably not that useful too in practice as students should still be able to read out the display when the screen is dimmed, assuming the normal brightness level is properly set in the first place

However, being able to set the delay before going to sleep is a good idea since students in exams might:
* prefer a longer timeout than normal because their calculator keeps turning off after thinking for a bit, especially if they are really stressed out ;
* prefer to dial down both brightness and timeout if they realize they are running out of battery at the worst possible time.

I think it'd be even more useful if the calculator could display the current time (to prevent the lost-track-of-time-because-of-no-visible-clock syndrome in exam), but that's not possible for now.

Now, I know themes support just got shot down, but as long as we have a new settings section, it would really not cost much to rename it from "Brightness" to "Appearance" and add a setting for inverting colors in this menu as a poor man's night mode. Just sayin'.

jperon:
Thank you all for your interest.

I had a request for those settings, not from a student, but from a teacher ; the main issue for him is that the screen gets dimmed too quick, and he doesn't see well with minimal brightness. So, for those who'd have the same expectation, the two "dim" settings would be most important.

Otherwise, I agree with @boricj about the renaming from "Brightness" to "Appearance" and color inverting.

boricj:
The firmware currently sets the brightness level to minimum when dimming the screen. Perhaps reducing the brightness instead (either by a set amount or a percentage of the normal level) would solve that problem. Smartphones don't set their brightness level to minimum when dimming the screen.

jperon:
That's one issue this PR solves. I added a setting for that, because expectations could be different : someone could want to prioritize battery saving while another one would like to see better.

quentinguidee:
Hello @jperon !

Can we reuse this pull request in [Omega](https://github.com/Omega-Numworks/Omega) and add your name in our contributors page ? Thank you :)

jperon:
@quentinguidee For sure you can reuse it, with or without my name.
Thanks !

quentinguidee:
Thank you :) !

RedGl0w:
I'll PR your branch to update with epsilon master today

PR #916, opened 2019-03-12
[graph] Values tab: select the right cell of the values table when displaying/undisplaying the derivative column

EmilieNumworks:


EmilieNumworks:
Duplicate

PR #927, opened 2019-03-21
Squircled App Icons

0b101:
I slightly modified the shape of the app icons to be more visually appealing and modern.

Squircle Example

Here is a template to use as a mask over existing icons: SquircleTemplate


Ecco:
HI @0b101 ! Thank you for your contribution 😄 Unfortunately we have no plan to change the look'n'feel, so I'll have to close this PR. Thanks again though!

PR #937, opened 2019-03-25
Add font size option

### LINKED ISSUE #400: Font size in Python editor too large
Is it possible to consider either a smaller font, or have a way to choose a smaller font in the settings, so you could get a better overview of the edited text?

### LINKED ISSUE #655: Long lines in Python editor
Hi, it would be nice to see directly whole (long) lines in Python editor. 
Yet it is hidden on right part, not so easy to read on calculator. A dedicated arrow on the left to indicate that it is not a new line but the end of the least one could be nice to use ? E.g. \u21b3 ↳

[Quite related with [#400 issue](https://github.com/numworks/epsilon/issues/400), for more or less the same reason]

Thanks

0b101:
**Please do not merge yet**
Fixes #400 and partially fixes #655 

#### Screenshots
Left is the old (**large**) font and right is the new (**small**) font
PythonBefore fontSmall


boricj:
Older teachers with less-than-perfect eyesight won't be pleased by this. At the very least, this should be an option and not a choice forced down their throats.

0b101:
> At the very least, this should be an option and not a choice forced down their throats.

Something like this?
Settings1 Settings2

0b101:
The preferences code seemed to change since the last time I added a new setting. Please don't merge this until I figure out how to pass KDFont through sharedPreferences.

Ecco:
Thank you @0b101 , that's definitely something we want to do! That being said, we had the exact same idea in mind as @boricj : the smaller font *has* to be an option, and the large font *has* to be the default value for this option 😄

0b101:
Should I make a full font size tab or a simple switch for changing modes?

0b101:
I should be about finished after I figure out how to get past this error:
`poincare/include/poincare/preferences.h:49:24: error: could not convert 'KDFont::SmallFont' from 'const KDFont* const' to 'KDFont'
         return KDFont::SmallFont;
                        ^~~~~~~~~
poincare/include/poincare/preferences.h:51:24: error: could not convert 'KDFont::LargeFont' from 'const KDFont* const' to 'KDFont'
         return KDFont::LargeFont;`

artaxxx:
Thanks @0b101 ! I agree with you, we'll add the option in the Settings app. It will be easier if we allow the user to change the font size in an other app in the future :)

0b101:
I'm closing this for the time being. If I ever get this working, I'll reopen this request.

Issue #587, opened 2018-06-26
Autocompletion


Salamandar:
Autocompletion would be great in Calculs/Python. I guess MicroPython provides such feature.

Maybe by a double-click on the right arrow ?

matheod:
Or an autocompletion menu in the toolbox ?

0b101:
It would be great for parentheses ( )

Salamandar:
> Or an autocompletion menu in the toolbox ?

That may be too "hidden" but that's an idea. Or maybe a long-click on the toolbox button to show the autocompletion list ? :)

UDXS:
Simply placing ')' when '(' is typed and then placing it between them would make programming faster. This should be done with '{', '[', and quotations.

Salamandar:
> Simply placing ')' when '(' is typed and then placing it between them would make programming faster. This should be done with '{', '[', and quotations.

That's not what I meant by autocompletion but that's a great start yeah ;)

XanderLeaDaren:
I support autocompletion, also for functions’ and variables’ names!

artaxxx:
Fixed in 14.4.0

PR #939, opened 2019-03-26
shift+( now displays ()

0b101:
This fixes part of #587.


artaxxx:
Thanks @0b101 . It's a little strange behavior. In my view, double parenthesis is a text editor feature. So the question would rather be to know if pressing the "( " key makes a double parenthesis directly or not. Going through the shift key seems incongruous to me.

On the other hand, it's not something we necessarily want to do. The problem you are solving here is an something about autocomplete. And I think that would be the right answer! So I'm closing this PR.

PR #940, opened 2019-03-26
Added brightness hotkeys

### LINKED ISSUE #920: Hot keys to change screen brightness
Looks like there is no ambient light sensor in the calculator. So, screen is either to bright or too dark, depending on environment lighting.

Need to have hotkeys for brightness. In other calculator I have «Shift + Left» and «Shift + Right» control the brightness. I see these combinations are used for other tasks in Numworks.

So, may be, make «Shift + Up» and «Shift + Down» control the brightness?

0b101:
This fixes #920 by adding shift++ (shift and plus) and shift+- (shift and minus) hotkeys to change the brightness.

Please note: the hotkeys do not update the brightness slider while in settings. To see the new location of the slider just go to any other page in settings (or home) then go back.

### DIFF
--- a/apps/apps_container.cpp
+++ b/apps/apps_container.cpp
@@ -220,6 +220,12 @@ bool AppsContainer::processEvent(Ion::Events::Event event) {
     suspend(true);
     return true;
   }
+  if (event == Ion::Events::BrightnessPlus || event == Ion::Events::BrightnessMinus) {
+      int delta = Ion::Backlight::MaxBrightness/GlobalPreferences::NumberOfBrightnessStates;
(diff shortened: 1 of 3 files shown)


nanoant:
Wow. This is great addition. Thanks!
For the slider not updating, we could potentially consume same event in `MainController::handleEvent` of `apps/settings/main_controller.cpp` where we do `if (model()->children(selectedRow())->label() == I18n::Message::Brightness) {` but unconditionally, so if it is active it will do both the brightness and slider adjustment otherwise the global even handler will fire?

PR #946, opened 2019-03-31
Display MicroPython version in settings

0b101:
MP


apps/settings/base.fr.i18n
```diff
@@ -20,6 +20,7 @@ Cartesian = "Algébrique "
 Polar = "Exponentielle "
 Brightness = "Luminosite"
 SoftwareVersion = "Version du logiciel"
+MicroPythonVersion = "Version du MicroPython"
```
adriweb:
```suggestion
MicroPythonVersion = "Version de MicroPython"
```

artaxxx:
Thanks! All things considered, we won't merge this PR. This information is non-essential and a bit too technical. Moreover, in my view, it shouldn't appear here: the version of python of my computer doesn't appear in this section for instance.

PR #947, opened 2019-03-31
Full-screen post-processing effects/Accessibility

boricj:
This is an experiment to see if post-processing drawing operations can have useful applications, especially regarding accessibility for visual impairments. The user-facing changes is a new "Accessibility" menu with two sample effects so far:
 * color inversion
 * 2x magnifier.

The magnification can be controlled by hitting Shift then a digit between 1 and 9 to zoom on another quadrant, but I'll gladly take suggestions on how to change this for something less clunky.

Under the hood, the NumWorks calculator doesn't have the RAM for a conventional approach to full-screen post-processing, so this uses fairly intricate bijective* transformations of low-level graphical primitives to compute effects on the fly. The net result is that post-processing effects can be linked into a pipeline that goes both ways in order for blending effects to work (their quality will depend on how lossless the transformations are) with hopefully acceptable run-time characteristics.

The reason I went with this approach is that it is completely transparent for the firmware. Adding effects requires implementing two-way transformation algorithms in a class, adding the hook for the pipeline and creating the UI for it. Effects can be disabled in some screens if needed, like the graphical output of a Python program or magnifying the main menu. It is also platform-agnostic, unlike previous inverted color mode implementations, so it should work everywhere and not just the device. I haven't tested on anything but the simulator though.

This is not finished or perfect (I haven't implemented reverse transformation of magnification yet and I'm having trouble bending escher to my will for example), but this should give an idea whether or not this is a viable approach. It won't solve all possible accessibility issues, however I believe it's a step on the right direction. Other possible filters are colorblind modes and global/per-channel gamma correction for example, but since I'm no accessibility expert I don't know what can or should be implemented nor how.

In short, it's cheap to implement, doesn't clash with the rest of the firmware, works on any platform and has no performance penalty if no effects are activated.

*_Not quite exactly, but let's not get bogged down in semantics._

image

image

image

image


0b101:
The "about" menu appears to be missing.
missing

boricj:
> The "about" menu appears to be missing.

That's covered in the "having trouble bending escher to my will" part of my description. I just barely got it to work well enough for a public preview and I'm way past my bed time, I'll deal with that later.

boricj:
I've implemented per-channel gamma correction. This is useful for several accessibility use cases (at least in theory), from colorblindness to boosting contrast (and happens to be also suitable as a makeshift night mode).
Note that while I can _see_ that I'm altering color balance, I can't know for sure that's actually helping for the target audience since I am not colorblind...

image
image
image
image

And yes, effects can still be stacked:
image

apps/settings/main_controller.cpp
```diff
@@ -177,7 +183,7 @@ int MainController::typeAtLocation(int i, int j) {
     return 1;
   }
 #ifdef EPSILON_BOOT_PROMPT
-  if (j == 7) {
+  if (j == 9) {
```
0b101:
```suggestion
  if (j == 8) {
```
Fix display issues

apps/settings/main_controller.cpp
```diff
@@ -201,7 +207,7 @@ void MainController::willDisplayCellForIndex(HighlightCell * cell, int index) {
     return;
   }
 #ifdef EPSILON_BOOT_PROMPT
-  if (index == 7) {
+  if (index == 9) {
```
0b101:
```suggestion
  if (index == 8) {
```
Fix display issues

boricj:
_Way_ bit-rotten. Also integrated inside Omega.

PR #954, opened 2019-04-08
Display more calculations

### LINKED ISSUE #935: Display more calculations on screen
#### Problem you'd like to fix
Currently, there can be only 3 calculations display on screen.

#### Screenshots
Here is what currently happens: 
SingleLine0

This is what the Ti-Nspire CX and HP Prime do:
SingleLine SingleLine2

#### Describe the solution you'd like
Calculations + answers that can fit on a single line should be displayed on a single line while longer calculations + answers should be displayed on two lines (like how all calculations + answers are currently displayed). I don't think this could be too complicated to implement.

#### Describe alternatives you've considered
There could also be a font size option (small - medium - large) #937?

0b101:
This adds support for single lined calculations and fixed #935.
moCalc mmmmm
This was developed by @PiGuy2 and @0b101 

Here is a ROM dump: [epsilon.bin](https://tiplanet.org/modules/archives/download.php?id=2244957)
Flash using: [webdfu](https://ti-planet.github.io/webdfu_numworks/)


PiGuy2:
As I mentioned in #935, there is still a small problem with the rendering, so this is not quite ready to merge.

apps/calculation/history_controller.cpp
```diff
@@ -150,7 +150,7 @@ KDCoordinate HistoryController::rowHeight(int j) {
   }
   Calculation * calculation = m_calculationStore->calculationAtIndex(j);
   App * calculationApp = (App *)app();
-  return calculation->height(calculationApp->localContext()) + 4 * Metric::CommonSmallMargin;
+  return calculation->height(calculationApp->localContext()); //+ 2 * Metric::CommonSmallMargin;
```
0b101: Remove comment
```suggestion
  return calculation->height(calculationApp->localContext());
```

apps/calculation/calculation.cpp
```diff
@@ -60,19 +60,44 @@ KDCoordinate Calculation::height(Context * context) {
...
+      }
     } else {
       KDCoordinate approximateOutputHeight = approximateLayout.layoutSize().height();
       KDCoordinate exactOutputHeight = exactLayout.layoutSize().height();
       KDCoordinate outputHeight = max(exactLayout.baseline(), approximateLayout.baseline()) + max(exactOutputHeight-exactLayout.baseline(), approximateOutputHeight-approximateLayout.baseline());
-      m_height = inputHeight + outputHeight;
+      KDCoordinate exactOutputWidth = exactLayout.layoutSize().width();
+      KDCoordinate approximateOutputWidth = approximateLayout.layoutSize().width();
+      KDCoordinate outputWidth = exactOutputWidth + approximateOutputWidth;
+      singleLine = outputWidth + inputWidth < maxWidth - 70;
```
0b101: This should fix the bug with 11^9
```suggestion
      singleLine = outputWidth + inputWidth < maxWidth - 40;
```

apps/calculation/history_controller.cpp
PiGuy2: This is fixed by 6b93fad

0b101:
> As I mentioned in #935, there is still a small problem with the rendering, so this is not quite ready to merge.

@PiGuy2 The rendering issues appear to be resolved in https://github.com/PiGuy2/epsilon/pull/6

adriweb:
Nice ! (Also it made [the news](https://tiplanet.org/forum/viewtopic.php?f=97&t=22625) on TI-Planet :P)
Can you rebase (and fix conflicts)? Probably some commits could be squashed together too.

0b101:
This commit will be compatible with master/11.0.0 (without conflicts) once the fork is synced/updated and https://github.com/PiGuy2/epsilon/pull/7 is applied.

PiGuy2:
This is now compatible with master, and tested pretty well!

0b101:
I posted a new ROM dump onto TI-Planet if anyone want's to test this out. 
ROM dump: [epsilon.bin](https://tiplanet.org/modules/archives/download.php?id=2244957)
Flash using: [webdfu](https://ti-planet.github.io/webdfu_numworks/)

0b101:
There is currently a display bug with the expanded view. It can be fixed by adding the following to lines 108-110: 
`if (expanded) {`
`      *memoizedHeight = *memoizedHeight + singleMargin;`
`}`

apps/calculation/calculation.cpp
```diff
@@ -66,20 +66,44 @@ KDCoordinate Calculation::height(Context * context, bool expanded) {
...
+      if (singleLine) {
+        *memoizedHeight  = (inputHeight >= outputHeight) ? inputHeight + singleMargin : outputHeight + singleMargin;
+      } else {
+        *memoizedHeight  = inputHeight + outputHeight + doubleMargin;
+      }
     }
     /* For all display output except ExactAndApproximateToggle, the selected
```
0b101:
```suggestion
     if (expanded) {
         *memoizedHeight = *memoizedHeight + singleMargin;
     }
     /* For all display output except ExactAndApproximateToggle, the selected
```

PiGuy2: I compiled it with this change, but I'm not sure what bug it is fixing.

0b101: There is a bug with the expanded view where it either doesn’t display at all or it displays incorrectly. Try entering something like `0.8/0.3` and then pressing the up arrow after executing. After the change, it should correctly display the fractional answer along side the decimal answer when selected.

0b101:
This is just a quick update: I am currently rewriting this to be less buggy and optional (with a display style setting of normal or condensed).

0b101:
I’m closing this so I can upload my new version (from Omega) tomorrow

Issue #117, opened 2017-09-10
[toolbox] Chemistry and other scientific constants in Toolbox


tjhorner:
As per suggestion of @adriweb (thanks!), I'm opening an issue on this.

I'm sure many high school students, such as myself, would find that scientific constants in the Toolbox prove helpful during chemistry or physics work, for example.

I've already implemented [chemistry constants](https://twitter.com/tjhorner/status/906790966924734464) in my [personal fork](https://github.com/tjhorner/kaleculator-os), and they've been proven to be helpful already!

Adding these in the toolbox is _perfect_, since it's [non-intrusive](https://core.telegram.org/tsi/feature_philosophy#3-complexity) for those who don't need it, but still easily accessible for those who do.

The question that remains is: what constants should be placed in the Toolbox? Right now on my fork, I only have Avogadro's number and the molar mass of the elements.

adriweb:
It is a good idea indeed, many other calcs have this feature (including when in exam mode).
I'd also like to insist a lot on displaying the unit of the constant too, as it is quite important.

Regarding what to have, here's a list of physics constants: http://hyperphysics.phy-astr.gsu.edu/hbase/Tables/funcon.html 
and one for chemistry constants: https://en.wikibooks.org/wiki/General_Chemistry/Constants (that said, both could be merged)
There are also useful stuff here that could be used for conversions: https://chem.libretexts.org/Reference/Units_and_Conversions/Physical_Constants

tjhorner:
Yes, displaying the proper unit for the constant would be important. The only reason I didn't on my fork was because I was too lazy to include two localisation strings: one for the value with units (displayed), and one for the value without the units (inserted into text field).

But yes, in science units are super important. As my chemistry teacher says, "no naked numbers!" :)

Zezombye:
I don't think the toolbox is the right place for this - a better way would be to add a new physics app, with constants and periodic table.

adriweb:
A new app, dedicated to physics/chemistry, possibly a periodic table, is yet another thing on its own, this isn't the right github issue for it (even though I agree it needs to happen at some point, definitely).
I do think there needs to be some kind of "global toolbox" available across apps or something, that said a good first step would be to have those constants available at least in the calc app. It's mostly there that it'd be used.

tjhorner:
[...]

PR #956, opened 2019-04-11
Add physics and chemistry constants + translation

ifndev:
(suggested by @adriweb and @tjhorner , see #117)

Implementation of physics and chemistry constants in the toolbox
It can be useful to avoid errors when copying them from the subject during an exam or a test.

It adds the following approximative values (based on the ones given in the French baccalaureate and exercise books):

_(physics)_
Speed of light
Gravitational Constant
Coulomb's Constant
Elementary charge
Mass of a nucleon
Mass of a proton
Mass of a neutron
Mass of an electron
Plank constant

_(chemistry)_
Avogadro constant

I decided to include both the general mass of a nucleon and the mass of a proton/neutron because i'm unsure about those given at the exams in different countries, it may be confusing tough.

It may be necessary to check translations

**EDIT**: I removed most of the specials characters in translations, but i figured out they were replaced automatically... I won't change anything for the moment, up to you to decide whether or not they should be deleted


0b101:
Here are some screenshots. I would recommend that you move the constants node lower.
C1C2
C3C4

boricj:
I would expect constants given by the calculator to be far more precise than that.

Lesquivemeau:
Well, i partly agree with you but unfortunately the values given at the bachalaureate are these ones

(after checking, it seems that the nucleon's approximate mass is the one used most of the time)

adriweb:
> Well, i partly agree with you but unfortunately the values given at the baccalaureate are these ones

I'm not sure how that's relevant though - if the user is given specific numbers to use he will input them directly. But if he wants to look them up and use them independently, he should be able to do so directly from the calc with greater precision.

From what I see on the Nspire CX CAS, constants are given with 10 sig. figs. (and with units built-in, which allows conversions)

Lesquivemeau:
Interesting idea, giving the choice to the user would be a great idea.
Do you think it would be more relevant in the calculator settings (in order not to bloat the toolbox) or in the toolbox ?

adriweb:
Ah, no, I wasn't clear enough apparently ^^

In my opinion, the calc should only provide greater-precision values because I think that users needing to use specific ones (given by exams or for any other reasons) would type them directly and not bother fetching them from the menus.

Lesquivemeau:
Well, it's a good point, but when in an exam copying the values from the subject can result in several errors. However, i don't think having a complete app dedicated to constants (like on the 83pce) would be a good idea, as it would be unintuitive to use

Lesquivemeau:
I will change actual values for more precise ones, but i strongly consider adding a menu to chose between precise and approximate values

apps/shared.universal.i18n
```diff
@@ -1,15 +1,14 @@
...
-c_G = "6.67*10^-11"
-c_k = "9.0*10^9"
-c_e = "1.60*10^-19"
-c_Na = "6.02*10^23"
-c_mg = "1.67*10^-27"
-c_mp = "1.673*10^-27"
-c_mn = "1.675*10^-27"
-c_me = "9.109*10^-31"
-c_h = "6.63*10^-34"
+c_c = "299 792 458"
```
0b101: "299 792 458" should be written as "299792458"
```suggestion
c_c = "299792458"
```

nsajko:
Adding these and other constants is a must for NumWorks because other calculators have them, even much cheaper ones; but, @Lesquivemeau, you can not just pick arbitrary "approximations" to well known constants. Every constant definition should be as close to the actual (defined or known) constant as possible, and for maintainability each constant definition should be accompanied by a reference (preferably URI), like here: https://github.com/golang/go/blob/54abb5ff868bf20239bd7960038a28c3f31ba25e/src/math/const.go#L10 .

For example, NIST and CODATA ( https://physics.nist.gov/cuu/Constants/index.html https://www.codata.org/committees-and-groups/fundamental-physical-constants ) have a different opinion of Avogadro's constant's definition than you ;) .

One thing that should be noted is that it is wrong to just copy constant definitions from some textbook or exam because there are many different textbooks and exams over the world.

sagessylu:
@ifndev There is a typo in `apps/toolbox.fr.i18n` :
`c_hDesc = "Constante de Plank"` should be replaced with
`c_hDesc = "Constante de Planck"`

RedGl0w:
>  @ifndev There is a typo in `apps/toolbox.fr.i18n` :
> `c_hDesc = "Constante de Plank"` should be replaced with
> `c_hDesc = "Constante de Planck"`

You can also use suggestion in pr for this (click on the line you want to change, click on the + on le left, and then there's a button in the top bar for adding suggestion)

apps/toolbox.fr.i18n
```diff
@@ -53,3 +56,12 @@ RandomAndApproximation = "Aléatoire et approximation"
...
 PrimeFactorDecomposition = "Décomposition en facteurs premiers"
+c_cDesc = "Vitesse de la lumiere"
+c_GDesc = "Constante de gravitation"
+c_kDesc = "Constante de Coulomb"
+c_eDesc = "Charge élémentaire"
+c_NaDesc = "Nombre d'Avogadro"
+c_mpDesc = "Masse d'un proton"
+c_mnDesc = "Masse d'un neutron"
+c_meDesc = "Masse d'un électron"
+c_hDesc = "Constante de Plank"
```
sagessylu, line 67: There is a typo : it should be `c_hDesc = "Constante de Planck"`

apps/toolbox.fr.i18n
```diff
@@ -53,3 +56,12 @@ RandomAndApproximation = "Aléatoire et approximation"
 RandomFloat = "Nombre décimal dans [0,1["
 RandomInteger = "Entier aléatoire dans [a,b]"
 PrimeFactorDecomposition = "Décomposition en facteurs premiers"
+c_cDesc = "Vitesse de la lumiere"
```
sagessylu, line 59: There is a typo : it should be `c_cDesc = "Vitesse de la lumière"`

PR #961, opened 2019-04-17
Improvements to the Python kandinsky module

Ecco:
* A color is now simply a tuple of 3 RGB values between 0 and 255 (e.g. red is `(255,0,0)`)
* The function `draw_string` now takes two extra optional parameters, text and background color
* Add a `fill_rect` function to quickly draw a rectangle
* Syntax errors are shown *before* switching to the drawing screen


boricj:
While we're here, we could use a `wait_vsync()` function inside `kandinsky` to prevent flicker when rendering animations/games on the device (my pong PoC is quite nauseating without it). It doesn't have to be exposed in the toolbox since it's not that useful otherwise.

Ecco:
That's indeed a good suggestion but I think this function should go in another module. Maybe somewhere alongside a `scankey` function?

boricj:
I'm not a fan of exposing internal module names as Python modules (end users currently think `kandinsky` is a funky name for proprietary NumWorks Python functions as they can't know nor they care why it's named that way), but that ship has sailed a _long_ time ago.

We can either spin `kandinsky` into a complete NumWorks platform module or we can stick to the current theme and have an `ion` module. I believe for the latter that end-users won't see a logic as of why some things are arbitrarily split across arbitrarily-named modules and they'll end up writing things like `from kandinsky, ion import *` to get that out of the way.

0b101:
It would be useful if color could also accept hex color values.

Issue #974, opened 2019-05-17
Better device designation for manual flash with dfu-util


Louson:
#### Problem you'd like to fix

The build target %_flash uses only the word "Flash". This can be dangerous if I have another device which has "Flash" in its name. Same for the interface. The target uses the interface 0, which is possibly not.

https://github.com/numworks/epsilon/blob/4dd7d1d792522645c306bac7e1a1820cf233c0cc/scripts/targets.device.mak#L41-L43

#### Describe the solution you'd like

Maybe use the `<vendorID>:<deviceID>` and eventually the serial number ?
`$ dfu-util -d 0483:a291 -S BwAjAAhRNjE4NzU3 -s 0x08000000:leave -D $<`

Louson:
Wouldn't it be more secure to also use the DFU suffix ?

https://www.usb.org/sites/default/files/DFU_1.1.pdf at annex B. DFU File Suffix.

At flash, dfu-util says the suffix will be mandatory in future versions.

Ecco:
We don't use `dfu-util` anymore 😄

PR #975, opened 2019-05-17
Secure manual flash

Louson:
Detect device using the vendorID:productID description instead of
name "Flash" which may be used by others devices.

Flash using the same descriptor instead of the default interface which can be
already used by an other device.

See issue #974


Ecco:
Good idea, but you overlooked a specific case: the device can either run our own DFU code or ST's. And both don't use the same VID/PID.

Louson:
Then it must be possible to look into a list if the VID:PID associated to the name Flash is valid, what do you think ?

Ecco:
Hi @Louson. Thank you for your contribution!

Unfortunately we don't use dfu-util anymore, namely because it doesn't allow sending data to arbitrary regions that are not exposed in the memory map.

We did add the feature suggested in this patch though 😄

PR #977, opened 2019-05-20
Fixed german special characters

critor:
Now that unicode characters are supported (thanks), when appropriate, replaced occurrences of the 'ae', 'oe', 'ue', and 'ss' archaic digrams with the proper characters.


EmilieNumworks:
Thank you @critor, it needed to be done.

PR #979, opened 2019-05-25
[ion/shared] events: fix arc-trig function names

adriweb:


LeaNumworks:
Hello,
Thank you very much for your PR!
These modifications are also done by the older PR #973, which has also some code cleaning, so I merged 973 instead.

adriweb:
Oops, yes, thanks!

PR #986, opened 2019-05-28
[build] Update targets.device.mak to flash ./build/device/epsilon.elf

RemyDCF:
Currently, when building the firmware for the device, epsilon.elf is created in ./build/device/epsilon.elf. If we want to flash this firmware on a real device, the user have to move the epsilon.elf file to the root project folder, which can lead to some confusions, as this step is undocumented on  [the Numworks documentation](https://www.numworks.com/resources/engineering/software/sdk/).
This pull request makes the `make epsilon_flash` command get the epsilon.elf directly in the ./build/device/ folder, so the user can use the command straight away, without any additional step.

### DIFF
--- a/scripts/targets.device.mak
+++ b/scripts/targets.device.mak
@@ -34,7 +34,7 @@
 	@echo "==============================="
 
 .PHONY: %_flash
-%_flash: %.bin
+%_flash: ./build/device/%.bin
 	@echo "DFU     $@"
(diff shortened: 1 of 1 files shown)


notandxor:
thx, works for me (build & flash linux)

Ecco:
Hi @RemyDCF ! You're definitely right. We've fixed this a bit differently since we now don't use dfu-util. Thanks for the patch though, even if we couldn't merge it!

PR #988, opened 2019-05-31
Increase the max number of iterations to approximate an integral

### LINKED ISSUE #987: Integration bug
Seems my original comment wasn't saved.  I discovered a bug in the integration of x^(1/3).  Integrating x^(1/3) from 0 to b, where b<=17 returns a result;  b>17 returns nothing.  Why would this be?
CM

RubenNumworks:
Fix #987

The `Poincare::IntegralNode::adaptiveQuadrature` method approximates recursively the integral of a given function, more and more accurately. It stops when
* either the approximation error is smaller than a given tolerance
* or the number of iterations reaches a given maximum, where it returns `NaN`.

At each iteration, if the approximation error is not satisfactory, it divides the integration interval in two equal pieces and calls itself on each subinterval, expecting that their respective errors will both be smaller than the initial tolerance.
If the bounds of the integral are X and Y and the maximal number of iterations is N, the smallest subinterval size looks like (X - Y) / 2^N.

The more a function varies quickly (such as x -> x^a with a in (0,1) in the neighborhood of 0), the smaller the subinterval must be.

### DIFF
--- a/poincare/include/poincare/integral.h
+++ b/poincare/include/poincare/integral.h
@@ -41,7 +41,7 @@ private:
     T integral;
     T absoluteError;
   };
-  constexpr static int k_maxNumberOfIterations = 10;
+  constexpr static int k_maxNumberOfIterations = 20;
 #ifdef LAGRANGE_METHOD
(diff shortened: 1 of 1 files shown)


RubenNumworks:
Thank you for reporting this issue. Here is a quick pull request that allows to compute such integrals on larger (but still limited) intervals. We will try to fix this issue more generally soon.

Issue #992, opened 2019-06-04
Simulator breaks with some unicode char


CarloMara:
#### Describe the bug
Simulator has some keys with missing text/symbols. This happens to coincide with some random Unicode symbols. After checking which Unicode Char were used in  `epsilon/ion/src/simulator/keyboard/fltkkbd.cpp` I can confirm that my system font has all the symbols needed.

#### Screenshots
![broken_unicode](https://user-images.githubusercontent.com/5103073/58879029-f732c880-86d4-11e9-8c92-559faf3bbfb0.png)

#### To Reproduce
1. Build epsilon with `make PLATFORM=simulator`
2. Run `./build/simulator/epsilon.elf`

#### Expected behavior
Arrows and other math symbols are displayed correctly

#### Environment
Tried both in Manjaro and on an Ubuntu 18:04. Same bug. Commit 23e1e39.

Have a nice day,
Carlo

CarloMara:
In the quest to debug this problem I tried to compile a small fltk program with a button and some unicode strings.
Lo and behold, it doesn't work.

In case you're wondering this is the program.
```
#include <FL/Fl.H>
#include <FL/Fl_Window.H>
#include <FL/Fl_Box.H>
#include <FL/Fl_Button.H>

int main(int argc, char **argv) {
    Fl_Window *window = new Fl_Window(340,180);
    Fl_Button *button = new Fl_Button(20,40,300,100,"\u2191");
    window->end();
    window->show(argc, argv);
    return Fl::run();
 }
```
This is the result on my box
![fltk_only_button](https://user-images.githubusercontent.com/5103073/58899043-42ac9d00-86fc-11e9-8b9e-3b84e948d709.png)


I think something is acting up in how fonts are managed, ether on fltk or on manjaro.
I'll keep you posted, but this looks to be a bug outside epsilon.

Cheers,
Carlo

RemyDCF:
This is clearly a bug outside epsilon, as it's working on macOS

CarloMara:
I've done some further troubleshooting. I'm gonna write the conclusions I reached, later you'll find more details later.

1. It appears that the font loaded by default on my system doesn't support Unicode. I tried to understand who or where this font is specified, with no luck. I have the suspicion that it's something hard-coded ether in X or in Xwayland. 
[...]

PR #996, opened 2019-06-06
Fix problem with Unicode Symbol on some linux distirbutions

CarloMara:
See Issue #992 for details. 
In short, some Linux distro do not pack by default fonts supporting the Unicode symbols needed for some buttons. This patch force FLTK to load every font available on the system and then search for a known good one (Arial).


Ecco:
Hi @CarloMara ! Thanks for your contribution! That being said, we've moved away from FLTK, so I won't be able to merge it. Sorry 😄

PR #1020, opened 2019-06-18
[apps] Force packing classes of objects contained in the Storage

EmilieNumworks:
If these classes are padded, we lose space in the Storage and the
Storage::checksum is computed on uninitialized values (corresponding to
the padding memory spaces)


apps/sequence/sequence.h
```diff
@@ -91,7 +93,7 @@ friend class SequenceStore;
       assert(conditionIndex >= 0 && conditionIndex < 2);
       m_initialConditionSizes[conditionIndex] = size;
     }
-
+#pragma pack(pop)
```
LeaNumworks: Should be 6 lines below?

PR #1025, opened 2019-06-21
Fix cursor navigation on utf8

LeaNumworks:
Fix Left/Right/Backspace on special UTF8 glyphs ('é' for instance)

### DIFF
--- a/escher/src/text_input.cpp
+++ b/escher/src/text_input.cpp
@@ -96,6 +96,23 @@ bool TextInput::removeEndOfLine() {
   return false;
 }
 
+bool TextInput::moveCursorLeft() {
+  if (cursorLocation() <= text()) {
+    assert(cursorLocation() == text());
+    return false;
+  }
+  UTF8Decoder decoder(text(), cursorLocation());
+  return setCursorLocation(decoder.previousGlyphPosition());
+}
+
+bool TextInput::moveCursorRight() {
+  if (UTF8Helper::CodePointIs(cursorLocation(), UCodePointNull)) {
+    return false;
+  }
+  UTF8Decoder decoder(cursorLocation());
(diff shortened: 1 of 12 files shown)


ion/src/shared/unicode/utf8_decoder.cpp
```diff
@@ -63,6 +63,33 @@ CodePoint UTF8Decoder::previousCodePoint() {
...
+    followingCodePoint = nextCodePoint();
+  }
+  m_stringPosition = resultGlyphPosition;
+  return resultGlyphPosition;
+}
+
+const char * UTF8Decoder::previousGlyphPosition() {
+  assert(m_stringPosition > m_string);
+  CodePoint previousCP = previousCodePoint();
+  if (m_stringPosition == m_string) {
```
EmilieNumworks: Why do you need this special case? Won't it work the same without?

escher/src/text_input.cpp
```diff
@@ -99,3 +99,22 @@ bool TextInput::removeEndOfLine() {
 bool TextInput::privateRemoveEndOfLine() {
   return contentView()->removeEndOfLine();
 }
+
+bool TextInput::moveCursorLeftRight(bool right) {
```
EmilieNumworks: Why would you factorize moveCursorLeft and moveCursorRight if their implementation is a total disjunction?

ion/src/shared/unicode/utf8_decoder.cpp
LeaNumworks: That's right :)

escher/src/text_input.cpp
LeaNumworks: That's right too ! It was not a disjunction when I started cleaning.

PR #1030, opened 2019-06-25
Fix parentheses fraction

LeaNumworks:
Some layouts were wrongly serialized into parsable strings.
For instance,
(1
---   -> ((1)/(2)) -> (1/2)
2) 

We now use special "System parentheses" to serialise, which would give **(** (1 **)** / **(** 2) **)** -> Unparsable.

These system parentheses are now also used to serialize superscripts.


poincare/src/nth_root_layout.cpp
```diff
@@ -148,42 +148,17 @@ void NthRootLayoutNode::deleteBeforeCursor(LayoutCursor * cursor) {
...
-  // Case: root(x,n)
-  if (m_hasIndex
-      && (const_cast<NthRootLayoutNode *>(this))->indexLayout()
-      && !(const_cast<NthRootLayoutNode *>(this))->indexLayout()->isEmpty())
-  {
+  if (m_hasIndex) {
+    assert((const_cast<NthRootLayoutNode *>(this))->indexLayout());
+    if ((const_cast<NthRootLayoutNode *>(this))->indexLayout()->isEmpty()) {
+      // Case: root(x,empty): Write "'SquareRootSymbol'('radicandLayout')"
+      return SerializationHelper::Prefix(this, buffer, bufferSize, floatDisplayMode, numberOfSignificantDigits, SquareRoot::s_functionHelper.name(), true, false);
```
EmilieNumworks: Why true, false? Should it be integers?

poincare/src/nth_root_layout.cpp
```diff
@@ -148,42 +148,17 @@ void NthRootLayoutNode::deleteBeforeCursor(LayoutCursor * cursor) {
...
-    return bufferSize-1;
-  }
-
-  numberOfChar += (const_cast<NthRootLayoutNode *>(this))->radicandLayout()->serialize(buffer+numberOfChar, bufferSize-numberOfChar, floatDisplayMode, numberOfSignificantDigits);
-  if (numberOfChar >= bufferSize-1) { return bufferSize-1; }
-
-  numberOfChar += SerializationHelper::CodePoint(buffer + numberOfChar, bufferSize - numberOfChar, ')');
-  buffer[numberOfChar] = 0;
-  return numberOfChar;
+  return SerializationHelper::Prefix(this, buffer, bufferSize, floatDisplayMode, numberOfSignificantDigits, SquareRoot::s_functionHelper.name());
```
EmilieNumworks: Can't you do the same case for "Case: squareRoot(x)" and  "Case: root(x,empty): Write "'SquareRootSymbol'('radicandLayout')"?

PR #1036, opened 2019-07-01
Global access to app

RubenNumworks:
The `app()` may now be accessed from Escher and from every `App`.
- In Escher, a global `app()` function is defined in [app.h](https://github.com/numworks/epsilon/blob/master/escher/include/escher/app.h) (actually in [container.h](https://github.com/numworks/epsilon/blob/master/escher/include/escher/app.h)) in the global namespace (Escher is not namespaced).
- In every `App` (except those in `Shared`), a global `app()` function is defined in app.h in the corresponding namespace, say `Graph` for `Graph::App`. Hence there is **no need to cast** the `app()` anymore.

Those `app()` functions are defined thanks to new static (it is indeed assumed that there is only one `(Apps)Container` instance) methods
  - `sharedContainer()` in the `Container` class of Escher and
  - `sharedAppContainer()` in the `AppsContainer` derived class.
More precisely, the Container class holds a static member variable `s_sharedContainer` that is initialized in the method `AppsContainer::run()`, that is after the construction of the `AppsContainer` and of all the `App`s. This means that **the `app()` functions must not be called during the construction of an `App`**.

The `Responder::app()` and `App::container()` methods have been removed.

From now on, an `App` and its members may be accessed directly through `app()` instead of passing pointers from constructors to constructors, up to the classes where they are needed.

### DIFF
--- a/apps/apps_container.cpp
+++ b/apps/apps_container.cpp
@@ -1,4 +1,5 @@
 #include "apps_container.h"
+#include "apps_container_storage.h"
 #include "global_preferences.h"
 #include <ion.h>
 #include <poincare/init.h>
@@ -51,6 +52,11 @@ static KDColor sPromptColors[] = {
   Palette::YellowDark};
 
 #endif
+
+AppsContainer * AppsContainer::sharedAppsContainer() {
+  static AppsContainerStorage appsContainerStorage;
+  return &appsContainerStorage;
+}
 
 AppsContainer::AppsContainer() :
   Container(),
--- a/apps/apps_container.cpp
+++ b/apps/apps_container.cpp
@@ -138,8 +138,8 @@ void AppsContainer::suspend(bool checkIfPowerKeyReleased) {
   resetShiftAlphaStatus();
   GlobalPreferences * globalPreferences = GlobalPreferences::sharedGlobalPreferences();
 #ifdef EPSILON_BOOT_PROMPT
-  if (activeApp()->snapshot()!= onBoardingAppSnapshot() && activeApp()->snapshot() != hardwareTestAppSnapshot() && globalPreferences->showPopUp()) {
-    activeApp()->displayModalViewController(&m_promptController, 0.f, 0.f);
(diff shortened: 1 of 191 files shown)


apps/apps_container.h
```diff
@@ -24,6 +24,9 @@
 
 class AppsContainer : public Container, ExamPopUpControllerDelegate, Ion::StorageDelegate {
 public:
+  static AppsContainer * sharedAppsContainer() {
+    return static_cast<AppsContainer *>(Container::sharedContainer());
+  }
```
EmilieNumworks: Remove because AppsContainerStorage::sharedContainer() can be used instead.
(don't forget to change #include "apps_container.h" to #include "apps_container_storage.h")

apps/apps_container.cpp
```diff
@@ -241,6 +241,7 @@ bool AppsContainer::switchTo(App::Snapshot * snapshot) {
 }
 
 void AppsContainer::run() {
+  setSharedContainer(this);
```
EmilieNumworks: Remove this. Instead, fill s_sharedContainer when building AppsConainerStorage in AppsContainerStorage::sharedContainer().

apps/home/app.h
```diff
@@ -15,6 +15,7 @@ class App : public ::App {
   };
   class Snapshot : public ::App::Snapshot, public SelectableTableViewDataSource {
   public:
+    Snapshot() { selectCellAtLocation(0, 0); }
```
EmilieNumworks: Avoid doing complex things at construction

RubenNumworks:
Finally, the `app()` function in the namespace of each app has been replaced by a static accessor, member of each `App` class. Similarly in Escher, `Container::activeApp()` has been made static and replaces the `app()` function in Escher.
The static member `s_sharedContainer` in `Container`, not needed anymore, has been removed.

apps/home/app.h
RubenNumworks: Has been reverted

apps/apps_container.h
RubenNumworks: Container::sharedContainer() and AppsContainerStorage::sharedContainer() have been removed. Only AppsContainer::sharedAppsContainer() is left.

PR #1056, opened 2019-08-26
Some corrections in Portuguese language

tiagop:
Corrections in some misspelled words in portuguese i18n files.


LeaNumworks:
Obrigada !

PR #1068, opened 2019-09-14
[ion/sdl] Linux support

boricj:
image

This adds support for Linux on the SDL platform, with OpenGL acceleration and software fallback. Build with `PLATFORM=sdl MODEL=linux make`.

There's some trickery for bundling and loading the background JPEG file since ELF files don't have a native mechanism to bundle resources, hopefully the workaround is acceptable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ include scripts/struct_layout/Makefile
 include scripts/scenario/Makefile
 include quiz/Makefile # Quiz needs to be included at the end
 
-objs = $(call object_for,$(src))
+objs += $(call object_for,$(src))
 .SECONDARY: $(objs)
 
 # Load source-based dependencies
--- a/ion/src/external/stb_image.h
+++ b/ion/src/external/stb_image.h
@@ -1,0 +1,7559 @@
+/* stb_image - v2.23 - public domain image loader - http://nothings.org/stb
+                                  no warranty implied; use at your own risk
+
+   Do this:
+      #define STB_IMAGE_IMPLEMENTATION
+   before you include this file in *one* C or C++ file to create the implementation.
+
+   // i.e. it should look like this:
+   #include ...
+   #include ...
+   #include ...
+   #define STB_IMAGE_IMPLEMENTATION
+   #include "stb_image.h"
+
+   You can #define STBI_ASSERT(x) before the #include to avoid using assert.h.
+   And #define STBI_MALLOC, STBI_REALLOC, and STBI_FREE to avoid using malloc,realloc,free
+
+
+   QUICK NOTES:
+      Primarily of interest to game developers and other people who can
+          avoid problematic images and only need the trivial interface
+
+      JPEG baseline & progressive (12 bpc/arithmetic not supported, same as stock IJG lib)
+      PNG 1/2/4/8/16-bit-per-channel
+
+      TGA (not sure what subset, if a subset)
+      BMP non-1bpp, non-RLE
+      PSD (composited view only, no extra channels, 8/16 bit-per-channel)
+
+      GIF (*comp always reports as 4-channel)
+      HDR (radiance rgbE format)
+      PIC (Softimage PIC)
+      PNM (PPM and PGM binary only)
+
+      Animated GIF still needs a proper API, but here's one way to do it:
+          http://gist.github.com/urraka/685d9a6340b26b830d49
+
+      - decode from memory or through FILE (define STBI_NO_STDIO to remove code)
+      - decode from arbitrary I/O callbacks
+      - SIMD acceleration on x86/x64 (SSE2) and ARM (NEON)
+
+   Full documentation under "DOCUMENTATION" below.
+
+
+LICENSE
+
+  See end of file for license information.
+
+RECENT REVISION HISTORY:
+
+      2.23  (2019-08-11) fix clang static analysis warning
+      2.22  (2019-03-04) gif fixes, fix warnings
+      2.21  (2019-02-25) fix typo in comment
+      2.20  (2019-02-07) support utf8 filenames in Windows; fix warnings and platform ifdefs 
+      2.19  (2018-02-11) fix warning
+      2.18  (2018-01-30) fix warnings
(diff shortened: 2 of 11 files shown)


Ecco:
Great, thanks!

I see an issue with this PR: it tries really hard to emulate what is done for other platforms, but I'm not sure it makes a lot of sense on Linux. Namely:

* It tries to embed an image decompression library. Definitely something we'd rather avoid as it makes for a large binary. Let's just rely on a system-provided libjpeg.
* It also embeds the JPEG image into the ELF file.

Both of those things make sense on platforms where users are used to downloading standalone binaries. But AFAIK, this is definitely not how things are usually done on Linux.

boricj:
If we rely on a system-provided libjpeg, we might as well rely on a system-provided libsdl too since we'll lose distribution portability anyway. Should I amend the pull request in that way?

The bit about embedding JPEG files into the program however is important: epsilon doesn't have to find files at runtime and can thus be run from anywhere on the file-system, regardless of the current working directory.

Ecco:
> If we rely on a system-provided libjpeg, we might as well rely on a system-provided libsdl too since we'll lose distribution portability anyway.

Actually that's what I thought too. But on a second thought, it might make sense to build libsdl because:

* It allows us to control which SDL version is used
* libjpg is pretty much ubiquitous, so it would make sense to depend on it. It's definitely not the case for libsdl.

>  Should I amend the pull request in that way?

Please don't since it's going to conflict with #1118 anyway.

> The bit about embedding JPEG files into the program however is important: epsilon doesn't have to find files at runtime and can thus be run from anywhere on the file-system, regardless of the current working directory.

Well, that's true. I tried to find some examples to get some "good practices". The only one I found was Blender. They seem to be shipping images as separate files but are depending on some external libraries.

```
$ ldd blender
linux-vdso.so.1 (0x00007fffb2af8000)
librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f83a83d4000)
libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f83a81b7000)
libGLU.so.1 => /usr/lib/x86_64-linux-gnu/libGLU.so.1 (0x00007f83a7f48000)
libGL.so.1 => /usr/lib/x86_64-linux-gnu/libGL.so.1 (0x00007f83a7cd6000)
libX11.so.6 => /usr/lib/x86_64-linux-gnu/libX11.so.6 (0x00007f83a7996000)
libXi.so.6 => /usr/lib/x86_64-linux-gnu/libXi.so.6 (0x00007f83a7786000)
libXxf86vm.so.1 => /usr/lib/x86_64-linux-gnu/libXxf86vm.so.1 (0x00007f83a757e000)
libXfixes.so.3 => /usr/lib/x86_64-linux-gnu/libXfixes.so.3 (0x00007f83a7378000)
libXrender.so.1 => /usr/lib/x86_64-linux-gnu/libXrender.so.1 (0x00007f83a716e000)
libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f83a6f6b000)
libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f83a6d67000)
libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f83a69c8000)
/lib64/ld-linux-x86-64.so.2 (0x00005595fad0d000)
libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f83a66c2000)
libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f83a64ab000)
libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f83a6129000)
libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f83a5eff000)
libxcb-dri3.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri3.so.0 (0x00007f83a5cfc000)
libxcb-present.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-present.so.0 (0x00007f83a5af9000)
libxcb-sync.so.1 => /usr/lib/x86_64-linux-gnu/libxcb-sync.so.1 (0x00007f83a58f0000)
libxshmfence.so.1 => /usr/lib/x86_64-linux-gnu/libxshmfence.so.1 (0x00007f83a56ee000)
libglapi.so.0 => /usr/lib/x86_64-linux-gnu/libglapi.so.0 (0x00007f83a54bf000)
libXext.so.6 => /usr/lib/x86_64-linux-gnu/libXext.so.6 (0x00007f83a52ad000)
libXdamage.so.1 => /usr/lib/x86_64-linux-gnu/libXdamage.so.1 (0x00007f83a50aa000)
libX11-xcb.so.1 => /usr/lib/x86_64-linux-gnu/libX11-xcb.so.1 (0x00007f83a4ea8000)
libxcb.so.1 => /usr/lib/x86_64-linux-gnu/libxcb.so.1 (0x00007f83a4c7e000)
libxcb-glx.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-glx.so.0 (0x00007f83a4a63000)
libxcb-dri2.so.0 => /usr/lib/x86_64-linux-gnu/libxcb-dri2.so.0 (0x00007f83a485e000)
libdrm.so.2 => /usr/lib/x86_64-linux-gnu/libdrm.so.2 (0x00007f83a464e000)
libXau.so.6 => /usr/lib/x86_64-linux-gnu/libXau.so.6 (0x00007f83a444a000)
libXdmcp.so.6 => /usr/lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007f83a4242000)
libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007f83a402c000)
```

Ecco:
Or we could simply use a static libjpeg build? It's installed by default alongside the dynamic version on [Debian](https://packages.debian.org/buster/amd64/libjpeg62-turbo-dev/filelist).

Ecco:
This is being merged into #1118 . Doing an actual `git-merge` was too much of a trouble so it won't show in the logs unfortunately...

Anyway, thanks @boricj for your help!

PR #1077, opened 2019-09-18
Include documentation within Epsilon's source code

Ecco:


Ecco:
Note that the documentation isn't perfectly up-to-date, but that's kind of the point of this PR.

Having the documentation alongside the sources will:
* Make it easier to keep them in sync
* Allow anyone to open a PR to improve/modify the doc

PR #1092, opened 2019-09-20
Some post version 12 cleaning

RubenNumworks:
- With the newly added plot types in the Graph app, the `Shared::CartesianFunction` class may now represent a function whose curve may be plotted in `Cartesian`, `Polar` or `Parametric` mode, the class name is thus not relevant anymore. We choose `ContinuousFunction`. Indeed, even though the considered functions are not generally continuous mathematically speaking, we can not simply name it Function, since such a class already exists: it is the base class of `CartesianFunction` and `Sequence`. The name `ContinuousFunction` should be understood as marking the difference with the Sequence class. The associated class `Graph::CartesianFunctionStore` is renamed accordingly.

- The following private nested classes are simply renamed to `RecordDataBuffer`, since they reside in different specific scopes:
  - `Shared::Function::FunctionRecordDataBuffer`
  - `Shared::CartesianFunction::CartesianFunctionRecordDataBuffer`
  - `Sequence::Sequence::SequenceDataBuffer`.

- Finally, in `Shared::ContinuousFunction` the implementations of the `numberOfActiveFunctionsOfType` and `activeRecordOfTypeAtIndex` methods are factorized using `numberOfModelsSatisfyingTest` and `recordSatisfyingTestAtIndex`.

- In passing, the `isAddEmptyRow`, `modelStore`and `editExpression`methods in the `ExpressionModelListController` class are simplified.

### DIFF
--- a/apps/shared/expression_model_store.cpp
+++ b/apps/shared/expression_model_store.cpp
@@ -43,12 +43,13 @@ void ExpressionModelStore::tidy() {
 int ExpressionModelStore::numberOfModelsSatisfyingTest(ModelTest test) const {
   int count = 0;
   int index = 0;
+  Ion::Storage::Record record;
   do {
-    ExpressionModelHandle * m = privateModelForRecord(recordAtIndex(index++));
-    if (m->isNull()) {
+    record = recordAtIndex(index++);
+    if (record.isNull()) {
       break;
     }
-    if (test(m)) {
+    if (test(privateModelForRecord(record))) {
       count++;
     }
   } while (true);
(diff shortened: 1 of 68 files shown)


apps/shared/expression_model_store.cpp
```diff
@@ -59,21 +59,22 @@ Ion::Storage::Record ExpressionModelStore::recordSatisfyingTestAtIndex(int i, Mo
   assert(0 <= i && i < numberOfModelsSatisfyingTest(test));
   int index = 0;
   int currentModelIndex = 0;
+  Ion::Storage::Record record;
   do {
-    Ion::Storage::Record r = recordAtIndex(currentModelIndex++);
-    ExpressionModelHandle * m = privateModelForRecord(r);
+    record = recordAtIndex(currentModelIndex++);
+    ExpressionModelHandle * m = privateModelForRecord(record);
     if (m->isNull()) {
```
LeaNumworks: assert(!m->isNull());

PR #1101, opened 2019-09-22
[doc] Build instructions: tell users about the two hardware versions

FrnchFrgg:
Trying to flash with the wrong model code can result in soft-bricked devices; explain how to build the correct version and how to recover if a wrong flash happened.

### DIFF
--- a/docs/build/index.md
+++ b/docs/build/index.md
@@ -62,10 +62,22 @@ make PLATFORM=simulator
 
 You can also update your NumWorks calculator easily. Note that you'll need to press the Reset button and that all data on your calculator will be lost.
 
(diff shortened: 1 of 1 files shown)


Ecco:
Thanks for this PR @FrnchFrgg ! That being said, we'd rather avoid adding docs/steps for the users when things can be figured out automatically.

In this case, the memory layout of N0100 and N0110 is different and is reported over DFU, so it's possible to handle things automatically.

Namely, I think it would be better to extend the `%_flash` targets to:
1. Make sure the correct device is connected (to avoid soft bricking)
2. Give some indications if a different model is connected (to help the user)

EmilieNumworks:
I'm closing this PR and opening the #1475 issue for the reasons @Ecco gave.

PR #1118, opened 2019-09-25
Remove FLTK

Ecco:
All simulator targets now use SDL. Note that we now use a custom-built SDL2 even for the web target.


boricj:
Did you add support for `MODEL=linux` to the SDL target as in #1068? It's the last platform stuck with the FLTK simulator.

Ecco:
> Did you add support for MODEL=linux to the SDL target as in #1068? It's the last platform stuck with the FLTK simulator.

Nope, at the moment it just doesn't build on Linux. Definitely something we'll need to add though.

PR #1121, opened 2019-09-25
Values controller of Graph app

EmilieNumworks:
- Display parametric function with only one column instead of two
- Fix a bug: when displaying a long float, avoid blank cell!


apps/graph/values/values_controller.cpp
```diff
@@ -41,6 +41,18 @@ ValuesController::ValuesController(Responder * parentResponder, InputEventHandle
   m_selectableTableView.setDelegate(this);
 }
 
+Shared::Hideable * ValuesController::hideableCellFromType(HighlightCell * cell, int type) {
+  if (type == k_notEditableValueCellType) {
+    Shared::HideableEvenOddBufferTextCell * myCell = static_cast<Shared::HideableEvenOddBufferTextCell *>(cell);
+    return static_cast<Shared::Hideable *>(myCell);
+  } else if (type == k_editableValueCellType) {
```
LeaNumworks: Remove else if case and replace with:
assert(type == k_editableValueCellType);
Shared::StoreCell * myCell = static_cast<Shared::StoreCell *>(cell);
return static_cast<Shared::Hideable *>(myCell);

?

apps/graph/values/values_controller.cpp
```diff
@@ -230,23 +213,38 @@ ViewController * ValuesController::functionParameterController() {
...
   Ion::Storage::Record record = recordAtColumn(columnIndex, &isDerivative);
   Shared::ExpiringPointer<ContinuousFunction> function = functionStore()->modelForRecord(record);
   Poincare::Context * context = textFieldDelegateApp()->localContext();
+  bool isParametric = function->plotType() == ContinuousFunction::PlotType::Parametric;
   if (isDerivative) {
-    return function->approximateDerivative(abscissa, context);
+    evaluationX = function->approximateDerivative(abscissa, context);
+  } else {
+    Poincare::Coordinate2D<double> eval = function->evaluate2DAtParameter(abscissa, context);
+    evaluationX = eval.x2();
```
LeaNumworks: Change the name of these variables?
evaluationX -> evaluationX2 (or evaluationY)
evaluationY -> evaluationX1 (or evaluationX)

apps/sequence/values/values_controller.h
```diff
@@ -14,12 +14,17 @@ class ValuesController : public Shared::ValuesController {
...
+  KDCoordinate columnWidth(int i) override;
+  KDCoordinate cumulatedWidthFromIndex(int i) override;
+  int indexFromCumulatedWidth(KDCoordinate offsetX) override;
   void willDisplayCellAtLocation(HighlightCell * cell, int i, int j) override;
   I18n::Message emptyMessage() override;
   IntervalParameterController * intervalParameterController() override {
     return &m_intervalParameterController;
   }
 private:
+  static constexpr KDCoordinate k_abscissaCellWidth = 100;
```
LeaNumworks: Could be factorised with k_ordinateCellWidth and Graph::ValuesController:: k_cellWidth in Shared::ValuesController
columnWidth and indexFromCumulatedWidth become simpler?

LeaNumworks: It was done later.

PR #1128, opened 2019-09-27
Fixes for Linux simulator build

boricj:
These changes are required to build and run the simulator on a Chrome OS Linux container.

Note that documentation for building on other platforms should be updated too.

### DIFF
--- a/build/platform.simulator.linux.mak
+++ b/build/platform.simulator.linux.mak
@@ -1,2 +1,2 @@
 TOOLCHAIN = host-gcc
-EXE = bin
+EXE = elf
--- a/docs/build/index.md
+++ b/docs/build/index.md
@@ -31,14 +31,19 @@ brew install armmbed/formulae/arm-none-eabi-gcc fltk freetype libpng pkg-config 
 Most of the required tools are available as apt packages:
 
 ```
-apt-get install build-essential git libfltk1.3-dev libfreetype6-dev libpng-dev pkg-config
+apt-get install build-essential git libx11-dev libxext-dev libfreetype6-dev libpng-dev libjpeg-dev pkg-config
 ```
 
-You'll also need to install the latest version of  and make it available in your $PATH:
+You'll also need to install the latest version of GCC and make it available in your $PATH:
 
 1. Download the [GCC toolchain distributed by ARM](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads). You should obtain a `gcc-arm-none-eabi-x-linux.tar.bz2` file.
 2. Decompress that file with `tar xvfj gcc-arm-none-eabi-*-linux.tar.bz2`
 3. Add the resulting folder to your $PATH. If you use bash, ``echo "export PATH=\$PATH:`find $(pwd)/gcc-arm-none-eabi-*-update/bin -type d`" >> ~/.bashrc`` should do what you need (you'll need to restart your terminal afterwards).
+
+Alternatively, on Debian 10 and later you can directly install a sufficiently modern cross-toolchain:
+```
+apt-get install gcc-arm-none-eabi binutils-arm-none-eabi
+```
 
 ## Retrieve the source code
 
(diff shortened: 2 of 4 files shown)


build/platform.simulator.linux.mak
```diff
@@ -1,2 +1,2 @@
 TOOLCHAIN = host-gcc
-EXE = bin
+EXE = elf
```
Ecco: Why is this needed?

docs/build/index.md
```diff
@@ -31,15 +31,20 @@ brew install armmbed/formulae/arm-none-eabi-gcc fltk freetype libpng pkg-config
 Most of the required tools are available as apt packages:
 
 ```
-apt-get install build-essential git libfltk1.3-dev libfreetype6-dev libpng-dev pkg-config
+apt-get install build-essential git libx11-dev libxext-dev libfreetype6-dev libpng-dev libjpeg-dev pkg-config
 ```
 
-You'll also need to install the latest version of  and make it available in your $PATH:
+You'll also need to install the latest version of GCC and make it available in your $PATH:
```
Ecco: Good catch 😄

docs/build/index.md
```diff
@@ -31,15 +31,20 @@ brew install armmbed/formulae/arm-none-eabi-gcc fltk freetype libpng pkg-config
...
 
 1. Download the [GCC toolchain distributed by ARM](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads). You should obtain a `gcc-arm-none-eabi-x-linux.tar.bz2` file.
 2. Decompress that file with `tar xvfj gcc-arm-none-eabi-*-linux.tar.bz2`
 3. Add the resulting folder to your $PATH. If you use bash, ``echo "export PATH=\$PATH:`find $(pwd)/gcc-arm-none-eabi-*-update/bin -type d`" >> ~/.bashrc`` should do what you need (you'll need to restart your terminal afterwards).
 
+Alternatively, on Debian 10 and later you can directly install a sufficiently modern cross-toolchain:
+```
+apt-get install gcc-arm-none-eabi binutils-arm-none-eabi
+```
+
```
Ecco: Actually I purposedly avoided using distro-provided packages because newcomers might get hit by a bug that wouldn't have been picked up by our CI pipeline.

And actually I'd like to make the whole thing even easier by providing a pre-made meta-package. Think `apt-get install numworks-sdk`.

ion/src/simulator/shared/main.cpp
```diff
@@ -77,7 +77,13 @@ void init() {
 
   SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "1");
 
+  // Try creating a hardware-accelerated renderer.
   sRenderer = SDL_CreateRenderer(sWindow, -1, SDL_RENDERER_ACCELERATED | SDL_RENDERER_PRESENTVSYNC);
+  if (!sRenderer) {
+    // Try creating a software renderer.
+    sRenderer = SDL_CreateRenderer(sWindow, -1, 0);
+  }
+  assert(sRenderer);
```
Ecco: I'm kind of wary about this because this code is ran on all platforms.
By the way, shouldn't we simply remove the `SDL_RENDER_ACCELERATED` flag? See https://wiki.libsdl.org/SDL_CreateRenderer#flags

build/platform.simulator.linux.mak
boricj: It's not mandatory, but since we generate ELF files under Linux we might as well use the correct file extension.

ion/src/simulator/shared/main.cpp
boricj: It fails too if `SDL_RENDERER_PRESENTVSYNC` is set. I would not be worried about the fallback since, well, it's a fallback and it will be very obvious if it is engaged: scaling is done with a nearest-neighbor algorithm. Better to have an ugly but working output than an assertion failure.

docs/build/index.md
boricj: On Debian Buster, we get version 7.3.1 of `arm-none-eabi-gcc`. It's not _that_ old, at least for now and it works fine on my end. I did mention it as an alternative to the preferred solution, whenever it stops working for epsilon we can amend that section.

Ecco:
Forwarded as #1137 . Our CI is configured with the `.bin` extension on Linux.

PR #1155, opened 2019-10-09
Calibrate SysTick

LeaNumworks:
millis() was quite false because we change the HCLK frequency in msleep.
This is mitigated by also changing the Systick frequency in msleep, but this provokes a small drift, visible in the Python sleep -> This method was changed to generate less frequency changes


ion/src/device/shared/drivers/timing.cpp
```diff
@@ -17,11 +17,16 @@ using namespace Device::Timing;
 void msleep(uint32_t ms) {
   // We decrease the AHB clock frequency to save power while sleeping.
   Device::Board::setClockFrequency(Device::Board::Frequency::Low);
+  // Change the systick frequency to compensate the KCLK frequency change
+  Device::Timing::changeSysTickFrequency(Ion::Device::Clocks::Config::HCLKLowFrequency);
```
EmilieNumworks: Call to `Device::Timing::changeSysTickFrequency` should be inside `Device::Board::setClockFrequency` implementation. As soon as the clock frequency is modified, systick has to be adjusted.

PR #1166, opened 2019-10-17
Changed de settings

RedGl0w:


RedGl0w:
Ok I made an error
Going to correct it tomorrow !

PR #1170, opened 2019-10-23
Give access to MathToolbox's model in third-party apps

boricj:
As part of my third-party RPN app (https://github.com/boricj/numworks-rpn), I provide a math toolbox to use in order to access advanced features. `MathToolbox` has the right tree model for me, but I need custom handling to fit the RPN input mechanism.

I see three easy options here:
- Change visibility of `Toolbox`'s `rootModel()` method (what I currently do: https://github.com/boricj/numworks-rpn/commit/d8ee77f28aa02a9c2e745ea063c97cbc5d1bb880)
- Put the tree model somewhere where I can access it from my third-party app's source code
- Carry a copy of the tree model in my app (I'd rather not)

This can also be an opportunity to revisit the Sequence app's own toolbox, which is based off `MathToolbox` and has a fairly convoluted implementation of its overrides.

This particular pull request is not meant to be taken as-is, but rather to discuss and decide what to do here.

### DIFF
--- a/apps/math_toolbox.h
+++ b/apps/math_toolbox.h
@@ -8,11 +8,11 @@ class MathToolbox : public Toolbox {
 public:
   MathToolbox();
 protected:
-  bool selectLeaf(int selectedRow) override;
-  const ToolboxMessageTree * rootModel() const override;
-  MessageTableCellWithMessage * leafCellAtIndex(int index) override;
-  MessageTableCellWithChevron* nodeCellAtIndex(int index) override;
-  int maxNumberOfDisplayedRows() override;
+  virtual bool selectLeaf(int selectedRow) override;
+  virtual const ToolboxMessageTree * rootModel() const override;
+  virtual MessageTableCellWithMessage * leafCellAtIndex(int index) override;
+  virtual MessageTableCellWithChevron * nodeCellAtIndex(int index) override;
+  virtual int maxNumberOfDisplayedRows() override;
   constexpr static int k_maxNumberOfDisplayedRows = 6; // = 240/40
 private:
   MessageTableCellWithMessage m_leafCells[k_maxNumberOfDisplayedRows];


EmilieNumworks:
I am not sure what you want to do exactly but it could be achieved by subclassing MathToolbox (or Toolbox) and overriding `rootModel()`, couldn't it?

boricj:
I haven't looked at this in a _long_ time, but the RPN bits require custom logic for the toolbox to work, specifically a custom `selectLeaf()` implementation (https://github.com/boricj/numworks-rpn/blob/master/rpn_toolbox.cpp#L16-L29). If I'm not mistaken, subclassing `MathToolbox` would work if its `selectLeaf()` method was declared `virtual` so that `NestedMenuController` would call the correct overload, but it is currently not.

I'll try this solution. If it works as expected, I'll update this PR and my RPN app.

boricj:
Huh. Turns out declaring a method virtual in a base class (`virtual bool NestedMenuController::selectLeaf(int selectedRow) = 0`) will affect an indirectly derived class (RPN's `Toolbox`), even if an ancestor class (`MathToolbox`) overrides the method without declaring it virtual. I did not expect that, but it appears to be indeed specified that way in the standards (https://en.cppreference.com/w/cpp/language/virtual).

I find it very confusing, so I've amended the PR to mark `MathToolbox::selectLeaf()` virtual. It does not actually do anything, but it makes it obvious what's going on. By the way, the source code of epsilon is not consistent on this: poincare has a whole lot of intermediate overrides marked virtual, but I've stumbled upon an example that is not, you may want to enforce a code style for this at some point.

The RPN app has been updated and will build against the master branch of epsilon now (https://github.com/boricj/numworks-rpn/commit/b395bdb1302e95945c430db1ca743b5bb4339bae).

EmilieNumworks:
The fact that in C++ you can do:

```
class Base {
protected:
  virtual int foo() const {
    return 2;
  }
};

class DerivedFromBase : public Base {
public:
  int foo() const override { return Base::foo();}
};


class DerivedFromDerived : public DerivedFromBase {
public:
  int foo() const override { return 4;}
};
```

without re-specifying `virtual int foo() const override` in `DerivedFromBase`  is used quite often throughout the code (for instance `Responder::didBecomeFirstResponder` is constantly overridden, same for `ViewController::viewWillAppear`). I would rather get rid of the unnecessary 'override' in Poincare instead of adding the `virtual` keyword everywhere.

EmilieNumworks:
I am closing this PR as it appears you can achieve the RPN app without any change. Another PR will be necessary if we want to homogenize the declaration of intermediary methods as `virtual` or not.

PR #1175, opened 2019-11-02
[ion/simulator] Add command-line arguments support

M4xi1m3:
Added arguments for screen-only mode, fullscreen and a usage menu, while still keeping the original functionality of the EPSILON_SDL_SCREEN_ONLY preprocessor variable.

### DIFF
--- a/ion/src/simulator/shared/main_sdl.cpp
+++ b/ion/src/simulator/shared/main_sdl.cpp
@@ -1,23 +1,50 @@
 #include "main.h"
 #include "display.h"
 #include "platform.h"
-#if !EPSILON_SDL_SCREEN_ONLY
 #include "layout.h"
-#endif
 
 #include <assert.h>
+#include <string.h>
 #include <ion.h>
+#include <stdio.h>
 #include <ion/timing.h>
 #include <ion/events.h>
 #include <SDL.h>
 #include <vector>
 
+static bool argument_screen_only = false;
+static bool argument_fullscreen = false;
+
 void Ion::Timing::msleep(uint32_t ms) {
   SDL_Delay(ms);
 }
 
+void print_help(char * program_name) {
+  printf("Usage: %s [options]\n", program_name);
+  printf("Options:\n");
+  printf("  -f, --fullscreen          Starts the emulator in fullscreen\n");
+  printf("  -s, --screen-only         Disable the keyboard.\n");
+  printf("  -h, --help                Show this help menu.\n");
+}
+
 int main(int argc, char * argv[]) {
   std::vector<const char *> arguments(argv, argv + argc);
+
+  for(int i = 1; i < argc; i++) {
+    if(strcmp(argv[i], "-h")==0 || strcmp(argv[i], "--help")==0) {
+      print_help(argv[0]);
+      return 0;
+    } else if(strcmp(argv[i], "-s")==0 || strcmp(argv[i], "--screen-only")==0) {
+      argument_screen_only = true;
+    } else if(strcmp(argv[i], "-f")==0 || strcmp(argv[i], "--fullscreen")==0) {
+      argument_fullscreen = true;
+    }
+  }
+
+#if EPSILON_SDL_SCREEN_ONLY
+  // Still allow the use of EPSILON_SDL_SCREEN_ONLY.
+  argument_screen_only = true;
+#endif
 
   char * language = IonSimulatorGetLanguageCode();
   if (language != nullptr) {
@@ -41,13 +68,9 @@ namespace Main {
 
 static SDL_Window * sWindow = nullptr;
 static SDL_Renderer * sRenderer = nullptr;
-#if !EPSILON_SDL_SCREEN_ONLY
 static SDL_Texture * sBackgroundTexture = nullptr;
-#endif
 static bool sNeedsRefresh = false;
-#if EPSILON_SDL_SCREEN_ONLY
 static SDL_Rect sScreenRect;
-#endif
 
 void init() {
   if (SDL_Init(SDL_INIT_VIDEO) != 0) {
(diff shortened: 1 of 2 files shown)


Ecco:
Hi @M4xi1m3 ! Interesting, but what's the point of this?

M4xi1m3:
The complete simulator (as without EPSILON_SDL_SCREEN_ONLY) has very bad scaling (at least on my PC). Using EPSILON_SDL_SCREEN_ONLY makes it easier to take screenshots, but having that as a preprocessor variable is, in my opinion, a bit problematic. You have to recompile the whole thing if you want to use it and the average user might not know how to compile epsilon (I know, there is the web simulator, but it's a bit of a pain to build and use). Plus, if you're with students in a classrooms doing python with them on a numworks, you don't need the keyboard. You can just project the whole screen (by launching the simulator with -s -f) and type with your keyboard. It  makes it easier for them to read.

Also, I really think the simulator should be distributed as binary form, not only as a web page.

Ecco:
The reason we have EPSILON_SDL_SCREEN_ONLY is because on the Web simulator we don't want to draw the calculator's keyboard using SDL. Making this a compile-time option allows us to strip down the build (for instance, we don't need an image decompression routine).

> Also, I really think the simulator should be distributed as binary form, not only as a web page.

This I strongly disagree with. Using the web-based version virtually has only upsides:
 - You don't need to install it (and to uninstall it)
 - It works on virtually any OS
 - It is sandboxed (and therefore doesn't need any "admin" rights)
 - Can very well be used offline too
 - Doesn't need code signing

And sure enough, we do have support for (native) screenshots and fullscreen mode on the web version.

The reason we have Linux/macOS/Windows simulators in the first place is because they're great debugging tools. We have no plan to distribute them to a wider audience at the moment given how superior the web version is.

> The complete simulator (as without EPSILON_SDL_SCREEN_ONLY) has very bad scaling (at least on my PC)

So maybe this is what we should consider fixing? Would you mind sharing your OS details?

ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -1,24 +1,55 @@
...
+      print_help(argv[0]);
+      return 0;
+    } else if(strcmp(argv[i], "-s")==0 || strcmp(argv[i], "--screen-only")==0) {
+      argument_screen_only = true;
+    } else if(strcmp(argv[i], "-f")==0 || strcmp(argv[i], "--fullscreen")==0) {
+      argument_fullscreen = true;
+    } else if(strcmp(argv[i], "-u")==0 || strcmp(argv[i], "--unresizable")==0) {
+      argument_unresizable = true;
+    }
+  }
```
nanoant: 1. Can you please keep formatting consistent with the rest of the code, i.e. space in `for (` and `if (` and spaces around ` == `.
2. Can you please use existing SDL factilities here instead of relying on libc, so `SDL_strcasecmp`, see `ion/src/simulator/external/sdl/src/test/SDL_test_common.c`.

ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -1,24 +1,55 @@
...
 #include "display.h"
 #include "platform.h"
-#if !EPSILON_SDL_SCREEN_ONLY
 #include "layout.h"
-#endif
 
 #include <assert.h>
+#include <string.h>
 #include <ion.h>
+#include <stdio.h>
```
nanoant: No need to include libc here. Use `SDL_strcasecmp` and `SDL_Log`.

ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -1,24 +1,55 @@
...
   SDL_Delay(ms);
 }
 
+void print_help(char * program_name) {
+  printf("Usage: %s [options]\n", program_name);
+  printf("Options:\n");
+  printf("  -f, --fullscreen          Starts the emulator in fullscreen\n");
+  printf("  -s, --screen-only         Disable the keyboard.\n");
+  printf("  -u, --unresizable         Disable resizing the window.\n");
+  printf("  -h, --help                Show this help menu.\n");
```
nanoant: Use `SDL_Log` instead.

ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -41,40 +72,43 @@ namespace Main {
...
+    sdl_window_args = SDL_WINDOW_ALLOW_HIGHDPI | SDL_WINDOW_FULLSCREEN;
+  }
+
+  if (argument_screen_only) {
+    sWindow = SDL_CreateWindow(
+      "Epsilon",
+      SDL_WINDOWPOS_CENTERED,
+      SDL_WINDOWPOS_CENTERED,
+      Ion::Display::Width, Ion::Display::Height,
 #if EPSILON_SDL_SCREEN_ONLY
```
nanoant: It is bit hard to follow what is happening there with `sdl_window_args` now. Can keep one `SDL_CreateWindow` call, but adjust `sdl_window_args` and `sdl_window_init_height/width` (some new var) instead?

ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -101,19 +135,27 @@ void relayout() {
...
-  SDL_Rect backgroundRect;
-  Layout::getBackgroundRect(&backgroundRect);
-
-  SDL_RenderCopy(sRenderer, sBackgroundTexture, nullptr, &backgroundRect);
-  SDL_RenderPresent(sRenderer);
-#endif
+  if (argument_screen_only) {
+    // Keep original aspect ration in screen_only mode.
+    float scale = (float)(Ion::Display::Width) / (float)(Ion::Display::Height);
+    if ((float)(windowHeight) * scale > float(windowWidth)) {
```
nanoant: Is this `(float)` casts necessary, or they're here just for readability, because scale is anyway float, thus `int` argument will be anyway implicitly converted to `float`. Same for two lines below.

EmilieNumworks:
Thanks @M4xi1m3 but for all reasons that @Ecco said, we won't maintain these features on Linux/macOS/Windows simulators.

Issue #1181, opened 2019-11-05
[Poincaré] With symbolic calculation enabled, some additions crashes the calculation app.


M4xi1m3:
#### Describe the bug
With symbolic computation enable, inputting something like (1)/(x^n) + m (n as real, m as any other expression), calculation app  crashes.

#### Screenshots
https://pastebin.com/24ngVqes

#### To Reproduce
1. Apply https://github.com/Omega-Numworks/Omega/commit/a196113cb0a97bff93d22d1f548d1600e475468d to get back symbolic computation
2. Go in the calculation app, with complex numbers disabled.
3. Type (1)/(x^2) + 3
4. Observe.

#### Expected behavior
It should just work as with complex numbers enabled.

#### Environment
 - Epsilon version: Latest master as time of writing (0736a3f), with symbolic computation enabled
 - The platform(s) on which the problem happens: Actual device (tested on n0110 and n0100), software simulator.

Ecco:
Then just don't apply this patch? 😄

Ecco:
Jokes aside, here's what happens:

1. It's actually not a crash. Poincare seems resilient enough to handle its own errors gracefully enough. If you wait long enough, you'll have a memory exhaustion popup.
2. For some reason we get stuck in a loop of 
```
Poincare::Addition::factorizeOnCommonDenominator(Poincare::ExpressionNode::ReductionContext)
Poincare::Addition::shallowReduce(Poincare::ExpressionNode::ReductionContext)
```

Thanks for the bug report @M4xi1m3

EmilieNumworks:
This is fixed by #1194. Thanks @M4xi1m3 !

PR #1194, opened 2019-11-13
[poincare] Power: always reduce (a^b)^(-1)

EmilieNumworks:
(a^b)^c -> a^(b*c)
This rule is not generally true: ((-2)^2)^(1/2) != (-2)^(2*1/2) = -2
This rule is true if:
- a > 0
- c is an integer

Warning 1: in real mode only c integer is not enough:
ex: ((-2)^(1/2))^2 = unreal != -2
We escape that case by returning 'unreal' if the a^b is complex.

Warning 2: If we did not apply this rule on expressions of the form
(a^b)^(-1), we would end up in infinite loop when factorizing an addition
on the same denominator.
For ex:
1+[tan(2)^1/2]^(-1) --> (tan(2)^1/2+tan(2)^1/2*[tan(2)^1/2]^(-1))/tan(2)^1/2
                    --> tan(2)+tan(2)*[tan(2)^1/2]^(-1)/tan(2)
                    --> tan(2)^(3/2)+tan(2)^(3/2)*[tan(2)^1/2]^(-1)/tan(2)^3/2
                    --> ...
Indeed, we have to apply the rule (a^b)^c -> a^(b*c) as soon as c is -1.

### DIFF
--- a/poincare/src/power.cpp
+++ b/poincare/src/power.cpp
@@ -625,18 +625,22 @@ Expression Power::shallowReduce(ExpressionNode::ReductionContext reductionContex
    *                     --> tan(2)+tan(2)*[tan(2)^1/2]^(-1)/tan(2)
    *                     --> tan(2)^(3/2)+tan(2)^(3/2)*[tan(2)^1/2]^(-1)/tan(2)^3/2
    *                     --> ...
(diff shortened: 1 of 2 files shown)


EmilieNumworks:
This fixed #1181

Issue #1192, opened 2019-11-12
Multiplication elision is inconsistent


Ecco:
#### Describe the bug
`2*x` is displayed as `2x` but `2*θ` is displayed as `2•θ`.

#### Screenshots
![screenshot (18)](https://user-images.githubusercontent.com/55989/68691271-c607cf80-0573-11ea-9370-49bca9fae662.png)

#### To Reproduce
1. Go to the 'Functions' app
2. Type `f(x)=2x`
3. Scroll and type `g(θ)=2θ`

EmilieNumworks:
Good one ! This is fixed by #1195.

PR #1195, opened 2019-11-13
[poincare] Fix multiplication symbol before Symbol Expression

EmilieNumworks:
This makes consistent 2x and 2θ (sign omission in both cases).

### DIFF
--- a/poincare/src/symbol.cpp
+++ b/poincare/src/symbol.cpp
@@ -107,6 +107,15 @@ Expression SymbolNode::shallowReplaceReplaceableSymbols(Context * context) {
   return Symbol(this).shallowReplaceReplaceableSymbols(context);
 }
 
+ExpressionNode::LayoutShape SymbolNode::leftLayoutShape() const {
+  UTF8Decoder decoder(m_name);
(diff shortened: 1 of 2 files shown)


poincare/src/symbol.cpp
```diff
@@ -107,6 +107,15 @@ Expression SymbolNode::shallowReplaceReplaceableSymbols(Context * context) {
   return Symbol(this).shallowReplaceReplaceableSymbols(context);
 }
 
+ExpressionNode::LayoutShape SymbolNode::leftLayoutShape() const {
+  UTF8Decoder decoder(m_name);
+  CodePoint codePoint = decoder.nextCodePoint();
+  if (codePoint == UCodePointNull || decoder.nextCodePoint() == UCodePointNull) {
```
LeaNumworks: Shouldn't it be : 
assert(codePoint != UCodePointNull);
if (decoder.nextCodePoint() == UCodePointNull) {
?
(+ take care of unused variable warning)

LeaNumworks:
Fixes #1192

PR #1201, opened 2019-11-15
Add an ion module in Python

Ecco:
At the moment it only provides a `keydown` function. Here's a small demo:

```python
from ion import *
if keydown(KEY_LEFT):
  print("Left key was pressed")
```


Ecco:
There's one gotcha to be aware of: on the simulator, `keydown` only works with the device's keyboard. For instance, pressing the left key on your PC's keyboard won't trigger a positive answer to `keydown(KEY_LEFT)`.

boricj:
Looks good, but I do think also providing a `get_keys()` like in my PR should be looked into. It enables one to use set operations, which is quite neat in practice:
```python
curkeys=get_keys()
if KEY_UP in keys:
  # UP key pressed down

oldkeys=keys
curkeys=get_keys()
if (not KEY_UP in oldkeys) and (KEY_UP in curkeys):
  # UP key just pressed down, simple version

curkeys=get_keys()-curkeys
if KEY_UP in curkeys:
  # UP key just pressed down, cooler version
```

Another minor problem with using numbers is that it's not very helpful when printed on the screen. Returning strings makes keys human-readable, which is easier for novice programmers to grasp:
```python
>>> get_keys()
{"EXE"}
```

Ecco:
The problem with using sets is that those are bulky, both in terms of memory and efficiency. Internally we use an `uint64_t` to represent a keyboard state, which is very quick to use (just a bit test), and very small in terms of memory.

And the problem with using strings is that those are error-prone : if we use named constants, we'll be able to warn a use who mistypes `KEY_LEFR`.

The only feature we don't address at the moment is the ability to scan once and test multiple times. I think it's a limitation that's really acceptable given how fast the scan really is. Anyway, and for the record, here's what a two-steps patch would look like.

```cpp
/* An Ion::Keyboard::State is 64 bits long.
 * We need to store it in a MicroPython object. We could use a multi-precision
 * integer, but those are harder to use. */

union StateOrFloat {
public:
  StateOrFloat(mp_float_t f) : m_f(f) {};
  StateOrFloat(Ion::Keyboard::State s) : m_s(s) {};
  mp_float_t floatingPoint() const { return m_f; }
  Ion::Keyboard::State state() const { return m_s; }
private:
  Ion::Keyboard::State m_s;
  mp_float_t m_f;
};

static_assert(sizeof(StateOrFloat) == sizeof(Ion::Keyboard::State));
static_assert(sizeof(StateOrFloat) == sizeof(mp_float_t));

mp_obj_t modion_keyboard_scan() {
  StateOrFloat sof(Ion::Keyboard::scan());
  return mp_obj_new_float(sof.floatingPoint());
}

mp_obj_t modion_keyboard_keydown(size_t n_args, const mp_obj_t * args) {
  mp_arg_check_num(n_args, 0, 1, 2, false);
  Ion::Keyboard::State state;
  if (n_args == 1) {
    state = Ion::Keyboard::scan();
  } else {
    assert(n_args == 2);
    StateOrFloat sof(mp_obj_float_get(args[1]));
    state = sof.state();
  }
  Ion::Keyboard::Key key = static_cast<Ion::Keyboard::Key>(mp_obj_get_int(args[0]));
  return mp_obj_new_bool(state.keyDown(key));
}
```

boricj:
I'm fine with `keydown(key)` returning a boolean, I'm fine with `keydown()` returning a set of integers or a set of strings, but I'm not a fan of `keydown()` returning a bitfield.

For integers vs strings, what I'm concerned about is exposing the raw hardware key mapping, which can create issues whenever the keyboard layout changes. We've already seen backwards-compatibility issues when `get_pixel()`/`set_pixel()` was changed from a R5G6B5 integer to a RGB24 tuple. Either through ignorance or minification people will write raw integers/strings and not use the identifiers.

While sets are less efficient than a plain integer, in practice the user will have at most 2-3 keys down at any time. I haven't measured the overhead, but I doubt it's going to be _that_ bad in practice. Besides, users will either busy-wait on `keydown()` or call it once per frame and do at most a couple of set operations, so it's not an area that requires utmost performance. Teachers and students are also likely to prefer using sets with intuitive operations than binary operators which may look like dark magic to some.

Ecco:
Makes sense. Let's just start with `keydown` returning a boolean then. That'll be a great first step 😄

mobluse:
Maybe one could be inspired by the button objects in Microbit:
https://microbit-micropython.readthedocs.io/en/latest/button.html
They have methods is_pressed(), was_pressed(), and get_presses().

PR #1225, opened 2019-11-30
Remove statically allocated buffers

boricj:
Following #1208, a bunch of static buffers in RAM were found that could be allocated on the stack or in Flash, most of them being rendering buffers. The calculator seems stable after these modifications. The `Poincare::Integer` static buffers have scary comments attached to them, so I left them alone (but I'm not quite convinced of their truthfulness).

Saves 3.5 KiB of RAM.

### DIFF
--- a/poincare/src/trigonometry_cheat_table.cpp
+++ b/poincare/src/trigonometry_cheat_table.cpp
@@ -82,7 +82,7 @@ Expression TrigonometryCheatTable::simplify(const Expression e, ExpressionNode::
  * For instance, when simplfy a Cosine, we always compute the value for an angle
  * in the top right trigonometric quadrant. */
 const TrigonometryCheatTable * TrigonometryCheatTable::Table() {
-  static Row sTableRows[] = {
+  static const Row sTableRows[] = {
     Row(Row::Pair("-90", -90.0f),
         Row::Pair("π*(-2)^(-1)", -1.5707963267948966f),
         Row::Pair("-100", -100.0f),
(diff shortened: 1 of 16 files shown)


poincare/src/trigonometry_cheat_table.cpp
```diff
@@ -82,7 +82,7 @@ Expression TrigonometryCheatTable::simplify(const Expression e, ExpressionNode::
  * For instance, when simplfy a Cosine, we always compute the value for an angle
  * in the top right trigonometric quadrant. */
 const TrigonometryCheatTable * TrigonometryCheatTable::Table() {
-  static Row sTableRows[] = {
+  static const Row sTableRows[] = {
```
quentinguidee:
```suggestion
  static Row sTableRows[] = {
```

I think the `const` makes crash some operations like `e^i.pi` or `cos(0)`

EmilieNumworks: I could not reproduce this!

EmilieNumworks:
I rebased your PR on master (#1471). You're probably right about the Poincare::Integer static buffer. But they at least guarantee that you can always do Integer computation when the stack is almost full so I am a bit cautious to change them now. It would require some extended testing...

poincare/src/trigonometry_cheat_table.cpp
quentinguidee: My bad, it was because of this : https://github.com/Omega-Numworks/Omega/commit/f763563834be83ffae397e047191a86fc59703b3#diff-5cf7de266c42fa98c531e899b90125e8 ; everything works perfectly

PR #1239, opened 2019-12-11
Exam mode upgrade

LeaNumworks:


boricj:
Wouldn't it be a better idea to store the exam mode inside the RTC backup registers? That wouldn't require odd Flash layouts.

M4xi1m3:
RTC backups registers seems to be cleared on reset on the n0110.

EmilieNumworks:
Hi @boricj,
Unfortunately, this would require the MCU vbat pin to be wired (which is currently not)... Also, RTC works on vbat mode only if clocked by a LSE source, which raises issues about consumption.

boricj:
Oh well, maybe for the next PCB revision then.

PR #1245, opened 2019-12-13
Fix gcd and lcm approximation

EmilieNumworks:
This fixes the following bug: when entering the sequence u(n) = gcd(n-2,n^2+n+3), in the value table, u(1) = -1 instead of 1.

### DIFF
--- a/poincare/src/approximation_helper.cpp
+++ b/poincare/src/approximation_helper.cpp
@@ -15,6 +15,18 @@ template <typename T> T absMod(T a, T b) {
   return result > b/2 ? b-result : result;
 }
 
+static inline int absInt(int x) { return x < 0 ? -x : x; }
+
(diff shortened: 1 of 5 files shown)


poincare/src/approximation_helper.cpp
```diff
@@ -15,6 +15,18 @@ template <typename T> T absMod(T a, T b) {
   return result > b/2 ? b-result : result;
 }
 
+static inline int absInt(int x) { return x < 0 ? -x : x; }
+
+template <typename T> int ApproximationHelper::IntegerApproximationIfPossible(const ExpressionNode * expression, bool * isUndefined, Context * context, Preferences::ComplexFormat complexFormat, Preferences::AngleUnit angleUnit) {
```
LeaNumworks: Rename to Absolute/Positive IntegerApproximationIfPossible?

PR #1250, opened 2019-12-17
[ion] Force inlining setBitRange

Ecco:


Ecco:
For some reason this PR does not build with older versions of GCC. The error looked like a bug:
```
ion/src/device/shared/boot/rt0.cpp:116: prohibited cross reference from .text.internal to `jump_to_external_flash() [clone .lto_priv.383]' in .text.external
````

Indeed, the linker script supposedly puts `jump_to_external_flash` in a custom section that is allowed to jump to .text.external.

It now builds thanks to 2c7a261.

Note that we *may* want to support older GCCs (and clang too) at some point. We'll need to figure out what was going on.

PR #1255, opened 2019-12-20
Clean use of L1 cache on N0110

EmilieNumworks:


ion/src/device/n0110/drivers/cache.cpp
```diff
@@ -77,23 +77,25 @@ void disableDCache() {
...
   isb();
 }
 
 void disableICache() {
   dsb();
   isb();
-  CORTEX.CCR()->setIC(false);
-  invalidateICache();
+  CORTEX.CCR()->setIC(false); // Disable I-cache
+  CORTEX.CCR()->setDC(true); // Invalidate I-cache
```
Ecco: ARGHHHH !!!!!

EmilieNumworks: Fixed!

PR #1258, opened 2019-12-27
Corrected graph Ymin/Ymax issue

RedGl0w:
When you try to set a window size, it doesn't include the bar size, and so the Ymin/Ymax is wrong.
I changed to include the bar size.

Before (I set 0 107 at Y window size) :
image
You can't see the 0 

After (I set 0 107 at Y window size):
image
You can see the 0

### DIFF
--- a/apps/shared/curve_view.cpp
+++ b/apps/shared/curve_view.cpp
@@ -110,7 +110,7 @@ const float CurveView::pixelWidth() const {
 }
 
 const float CurveView::pixelHeight() const {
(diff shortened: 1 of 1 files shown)


artaxxx:
Thanks @RedGl0w ! This behavior is a voluntary choice, so we're not going to merge this PR. The idea was to choose between two behaviors:
* the graph view goes to the bottom of the screen 
* the graph view goes to the top of the banner view
As the banner view can have different sizes (from 0 lines to two or three lines), we solved the question taking the most stable behavior.

RedGl0w:
I knowed, but I didn't continue working on it. But it is still an issue I think

PR #1261, opened 2020-01-01
Add the ability to change font size for the python app

MixedMatched:
Based on PR #937 by @0b101.

Also refactors main_controller to use labels instead of indices to identify s_model.

I'm unsure of translations and capitalization conventions (Python is a proper noun, but I'm not fluent in any language besides English.)


EmilieNumworks:
Hi @MixedMatched!
Thank you for your PR, this is indeed a great idea. So great that we have already developed it and it will be part of the next version! But thank you very much for your involvement!

PR #1268, opened 2020-01-07
Fix scenario creator

LeaNumworks:


Ecco:
>	$(Q) make -j8 PLATFORM=simulator clean && make -j8 DEBUG=1 ARCH=x86_64 PLATFORM=simulator epsilon.headless.bin

I think it would be better to get rid of the "ARCH=x86_64" part.

PR #1275, opened 2020-01-16
[build] Don't invoke bash-specific '[['

gkelly:
'[[' isn't available when $SHELL isn't set to bash, so use expr from
coreutils to check for the presence of the device in DFU mode.

$SHELL doesn't come from the environment, so I _think_ that on most Debian based distros this will always be /bin/sh, which isn't bash, inside of the make environment.


gkelly:
> Why are you using a regex match here? Seems like it would be better to use test and =.

Seemed more readable than:

```
if test "${DFU_SLAVE#*0483:df11}" != "$DFU_SLAVE"; then
```

jacobly0:
Oh sorry, I missed the * in the original for some reason.

EmilieNumworks:
Thanks @gkelly. I'd like to merge this PR but I can't if you don't sign the CLA.

busybox11:
bump

PR #1296, opened 2020-02-06
Python code editor: Shift+Up and Shift+Down can now be used to quickly navigate through the code

carl-vbn:
This is a very small feature that I always wanted.
When writing longer scripts in the Python app, it's very annoying to scroll all the way to the top.
Shift+Right and Shift+Left can already be used to jump to the beginning/end of a line but such a feature did not exist for Up/Down jumping. With this PR, you can jump to the top of your script with Shift+Up or to the bottom with Shift+Down.
I had to implement the ShiftUp and ShiftDown Ion Events for this to work.

### DIFF
--- a/escher/src/text_area.cpp
+++ b/escher/src/text_area.cpp
@@ -110,6 +110,10 @@ bool TextArea::handleEvent(Ion::Events::Event event) {
     contentView()->moveCursorGeo(-INT_MAX/2, 0);
   } else if (event == Ion::Events::ShiftRight) {
     contentView()->moveCursorGeo(INT_MAX/2, 0);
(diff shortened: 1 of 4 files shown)


RedGl0w:
Shift left is now used

EmilieNumworks:
Indeed, as @RedGl0w said, I am sorry but shift+[left|rigth|up|down] are now used to select text. Maybe, we could speed up the cursor navigation when pressing [left|rigth|up|down] for a "longer" time?

carl-vbn:
We'll that's actually great! Text selection is a feature I always wanted so I'll definitely be looking forward to that. And yeah pressing the arrow keys for a long time could indeed be used.

PR #1304, opened 2020-02-14
Improve telemetry

Ecco:


liba/src/strlcat.c
```diff
@@ -0,0 +1,19 @@
+#include <string.h>
+
+size_t strlcat(char * dst, const char * src, size_t dstSize) {
+  const size_t srcLen = strlen(src);
+  size_t dstLen = strlen(dst);
+  if (dstLen > dstSize) {
```
LeaNumworks: Replace these two conditions by
 if (dstLen >= dstSize) {
     return dstSize+srcLen;
 }

LeaNumworks: And this way, dstLen can become const

Issue #1325, opened 2020-02-20
Add a new backgroud and text color argument with Kandinsky draw_string


le-grand-mannitout:
Hi Numworks team ! I often would like to set a different color to the text or to the background, with the function "draw_string()".

The Kandinsky draw_string function is verry nice, but it could be well better by adding these 2 news arguments : this is currently quite difficult to do a nice graphic interface, without this functionality..

Here is a potential syntax of the function with this modification :
`draw_string("text", x, y, color_text, color_background)`

Thanking you in advance,

RedGl0w:
https://github.com/numworks/epsilon/blob/80c03220dd10ddf1d2bdadc55b240b8a4dc03c5e/python/port/mod/kandinsky/modkandinsky.cpp#L71
This already exists
It's just not in the toolbox
I advise you to rename your issue to `Add the colorbackground parameter in the toolbox of drawstring in kandinsky python module` or smth like this...

LeaNumworks:
Hello @le-grand-mannitout,
Thanks for creating this issue!
As @RedGl0w said, you can already use the `draw_string` method with those additional arguments, just like you suggested: `draw_string("text", x, y, color_text, color_background)`
We decided not to mention these arguments in the toolbox though, because a clear explanation would not fit in the toolbox space.

PR #1326, opened 2020-02-20
Added the 2 colors to the toolbox

RedGl0w:
This patches https://github.com/numworks/epsilon/issues/1325#issuecomment-589002116

I know this isn't perfect, but we need it to be added to the toolbox....


LeaNumworks:
Hi @RedGl0w, thanks for your PR!

We had indeed thought about adding those arguments in the toolbox, but decided not to.
We would like clear argument names, such as text_color and background_color, and these would not fit in the toolbox (on both the title and the definition lines).

By the way, color1 and color2 do not fit either
Screenshot 2020-02-20 at 16 53 02

I am thus closing the PR, but thanks again.

RedGl0w:
You should start thinking about a solution to make the text pass when it's too big in toolbox, because solving an issue by not solving another one is a bit dumb, I find...

PR #1328, opened 2020-02-20
[python] Add the choice of the python font size in draw_string()

RedGl0w:
## Demos
```py
draw_string("abc",150,150,color(0,0,0),color(255,255,255),C)
```
Result with `C = 1` or `C = True` :
unknown

Result without C, with `C = 0` or `C = False` :
unknown (1)

## Todo

- [x] Add in toolbox (removed by https://github.com/numworks/epsilon/pull/1326#issuecomment-589153962)


Ecco:
Hi  @RedGl0w ! That's a great suggestion! However using a boolean value makes little sense for the user here. We'd rather use a named constant like in `modion`.

RedGl0w:
> Hi  @RedGl0w ! That's a great suggestion! However using a boolean value makes little sense for the user here. We'd rather use a named constant like in `modion`.

The best solution could be to add in Kandinsky a new type : font
So we could use it everywhere when drawing string, but also for knowing which font is enable (this can be useful for menu in console)

Ecco:
>  knowing which font is enable (this can be useful for menu in console)

The console isn't handled by Kandinsky at all though… We could consider adding a `modsys` module, but that's quite out of the scope of this PR.

RedGl0w:
> > knowing which font is enabled (this can be useful for menu in console)
> 
> The console isn't handled by Kandinsky at all though… We could consider adding a `modsys` module, but that's quite out of the scope of this PR.

I didn't say that the actual font activated should be in kandisnky, I said that the font object should be in kandinsky, because kandinsky is where the fonts are.

PR #1333, opened 2020-02-21
[apps/code] Deduplicate definitions of pi and e

nsajko:
Merge the language-specific definitions of these mathematical constants
into one definition. Note that the definitions were in fact inconsistent
across languages. Also added a few decimals of precision to the
definitions, I guess the couple of bytes are not an issue.


LeaNumworks:
Hi @nsajko,
Thanks for this PR and for spotting the inconsistency!
However, we'd rather not have more precision in the toolbox definition than what Python provides. Could you please remove the decimals you added?
screenshot-4

nsajko:
@LeaNumworks OK, done.

LeaNumworks:
Great, thanks!

PR #1338, opened 2020-02-21
[python] Added record module

RedGl0w:
# Warning

This module is WIP

## What does it do ?

It adds a module to add a "file" system python module based on the Ion::Storage::Records, with 5 commands :

- `save(name,data)`
- `getdata(name)`
- `istaken(name)`
- `destroy(name)`
- `listrecords()`

## To do

- [ ] Patch the \x0 characters
- [ ] Patch the `listrecords()` issue with long records name
- [ ] Add to toolbox

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -140,6 +140,8 @@ port_src += $(addprefix python/port/,\
   mod/turtle/modturtle_table.c \
   mod/turtle/turtle.cpp \
   mphalport.c \
+  mod/record/modrecord_table.c\
+  mod/record/modrecord.cpp\
 )
 
 # Workarounds
--- a/python/port/genhdr/qstrdefs.in.h
+++ b/python/port/genhdr/qstrdefs.in.h
@@ -116,6 +116,15 @@ Q(time)
 Q(sleep)
 Q(monotonic)
 
+
+// record QSTRs
+Q(record)
+Q(istaken)
+Q(save)
+Q(getdata)
+Q(destroy)
+Q(listrecords)
+
 // MicroPython QSTRs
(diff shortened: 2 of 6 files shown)


Ecco:
Hi @RedGl0w ! That's an interesting suggestion, but we're not going to merge this because we'd rather stick to standard Python APIs whenever possible. In this case, that would be `open()` and `read()`.

RedGl0w:
> Hi @RedGl0w ! That's an interesting suggestion, but we're not going to merge this because we'd rather stick to standard Python APIs whenever possible. In this case, that would be `open()` and `read()`.

I'm currently working on that !

Ecco:
> I'm currently working on that !

Great! Note that I'm not really sure if/how we'd want to merge this though. There are tons of things that could go wrong using this kind of features, so maybe it'd be better to avoid them entirely.

bmuessig:
I would second the feature request for file IO. It can be quite useful for preserving state. Python isn't available in exam mode anyways, so you shouldn't have to worry about that.

adriweb:
Python may not be available in NL exam mode, but it is otherwise (standard exam mode)

RedGl0w:
This implementation, when it will be finished (so without a specific module) shouldn't be usable in exam mode.

bmuessig:
I've skimmed the storage records implementation and it seems that the file system global and uses CRC32 for file identification. Therefore, hash collisions are possible between user (python) files and system files. It might make sense to segregate python files from the other system files, if this was to be implemented. Otherwise an (unintentional) hash collision with a system file could silently break other apps.

PR #1343, opened 2020-02-21
Performance fixes relating to floating point operations

nsajko:
This replaces unnecessary double-precision soft-float operations with
single-precision floating-point operations, mainly by casting.

In a couple places I also replace a function call with a constant.

### DIFF
--- a/liba/src/external/openbsd/s_erff.c
+++ b/liba/src/external/openbsd/s_erff.c
@@ -1,0 +1,185 @@
+/* s_erff.c -- float version of s_erf.c.
+ * Conversion to float by Ian Lance Taylor, Cygnus Support, ian@cygnus.com.
+ * Taken from openlibm at 5b0e7e981321687ac0abe711fdeb30adcd9da932 and
+ * modified for Numworks Epsilon by Neven Sajko <nsajko@gmail.com>
+ */
+
+/*
+ * ====================================================
+ * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ *
+ * Developed at SunPro, a Sun Microsystems, Inc. business.
+ * Permission to use, copy, modify, and distribute this
+ * software is freely granted, provided that this notice
+ * is preserved.
+ * ====================================================
+ */
+
+//__FBSDID("$FreeBSD: src/lib/msun/src/s_erff.c,v 1.8 2008/02/22 02:30:35 das Exp $");
+
+#include "math.h"
+
+#include "math_private.h"
+
+static const float
+tiny	    = 1e-30,
+half=  5.0000000000e-01, /* 0x3F000000 */
+one =  1.0000000000e+00, /* 0x3F800000 */
+two =  2.0000000000e+00, /* 0x40000000 */
+/*
+ * Coefficients for approximation to erf on [0,0.84375]
+ */
+efx =  1.2837916613e-01, /* 0x3e0375d4 */
+efx8=  1.0270333290e+00, /* 0x3f8375d4 */
+/*
+ *  Domain [0, 0.84375], range ~[-5.4446e-10,5.5197e-10]:
+ *  |(erf(x) - x)/x - p(x)/q(x)| < 2**-31.
+ */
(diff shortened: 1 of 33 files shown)


nsajko:
There are some (two I think) cases involving overloaded functions where double precision is used for a reason I can not grasp yet. (I guess I am not that good with C++.) But let that be a future issue.

nsajko:
Can I get help with the failing test? The iOS build is failing, but I am not being shown why...

nsajko:
An accuracy test is failing on macos, but that probably just means the test should be fixed. The particular example in the test seems OK accuracy-wise, and indeed all the changes in this PR should be, as they basically just change the code to work as intended (use floats when supposed to, instead of using doubles just for some intermediate calculations).

Note that I have nothing against using doubles when necessary for greater accuracy, but at least most of the changes here are obviously a good trade-off. For the future: it would probably be a good idea to think of a way to make using soft-float double arithmetic more explicit.

nsajko:
Also, it would be good if you Numworks people could point me to some kind of document that explained your expectations regarding when to use double, and when to use float. Maybe you could make a whitelist of all the places where double may be used?

nsajko:
@LeaNumworks 

Turtle::drawPaw (python/port/mod/turtle/turtle.cpp) uses `float`s, and this seems incongruent with the fact that Epsilon's MicroPython is configured to use `double`s. Should I change one or the other?

liba/src/external/openbsd/s_logbf.c
```diff
@@ -0,0 +1,42 @@
+/* s_logbf.c -- float version of s_logb.c.
```
LeaNumworks: Hello,
We prefer to use OpenBSD 4.9's libm version (see 3adc5d9fac733fe04a4), and also not modifying the external files. Could you please replace this file with [s_logbf.c.zip](https://github.com/numworks/epsilon/files/4254640/s_logbf.c.zip)?

liba/src/external/openbsd/s_erff.c
```diff
@@ -0,0 +1,185 @@
+/* s_erff.c -- float version of s_erf.c.
```
LeaNumworks: Same comment as for s_logbf.c, could you please change this file with [s_erff.c.zip](https://github.com/numworks/epsilon/files/4254712/s_erff.c.zip)?

LeaNumworks:
> @LeaNumworks
> 
> Turtle::drawPaw (python/port/mod/turtle/turtle.cpp) uses `float`s, and this seems incongruent with the fact that Epsilon's MicroPython is configured to use `double`s. Should I change one or the other?

I see your point! However, I do not think it is necessary to change to `double`, as we are doing computations that should be fast and do not need precision.

LeaNumworks:
> An accuracy test is failing on macos, but that probably just means the test should be fixed. The particular example in the test seems OK accuracy-wise, and indeed all the changes in this PR should be, as they basically just change the code to work as intended (use floats when supposed to, instead of using doubles just for some intermediate calculations).
> 
> Note that I have nothing against using doubles when necessary for greater accuracy, but at least most of the changes here are obviously a good trade-off. For the future: it would probably be a good idea to think of a way to make using soft-float double arithmetic more explicit.

I pushed a commit to change the precision of this test, it should pass now. (While at it, I fixed the precision of some surrounding tests too).

liba/src/external/openbsd/s_logbf.c
nsajko: Sadly, that logbf (and logb) is wrong for subnormal inputs (it returns a constant). Could you import just this file from openlibm?

I started a discussion on the OpenBSD misc mailing list regarding this bug: https://marc.info/?l=openbsd-misc&m=158272385126733&w=2

LeaNumworks:
Hi @nsajko, sorry for the delay!
As stated in #1370, we've decided we'd rather use OpenBSD. I created PR #1486, which rebases this one and uses the OpenBSD files.
I will close this PR and merge the rebased one.
Feel free to open issues if you see any bug resulting from the OpenBSD faulty files, and thank you for emailing OpenBSD about the problem!

PR #1344, opened 2020-02-21
[poincare] Fix checking for special tree identifier value

nsajko:
Unsigned types are never negative.

### DIFF
--- a/poincare/include/poincare/tree_handle.h
+++ b/poincare/include/poincare/tree_handle.h
@@ -128,7 +128,7 @@ protected:
   void setIdentifierAndRetain(uint16_t newId);
   void setTo(const TreeHandle & tr);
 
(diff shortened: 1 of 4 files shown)


LeaNumworks:
Hello,
Thanks for spotting the issue. I would rather suggest creating a static `TreeNode` method
`static bool IsValidIdentifier(uint16_t id) { return  id < NoNodeIdentifier; }`
that you could use instead of your change, and also use it in `TreeHandle::hasNode`.
There are also some bad `m_currentIndex >= 0` checks that could be removed from tree_pool.h, if you want to change them :)

nsajko:
@LeaNumworks Done.

LeaNumworks:
Great, thank you!

PR #1345, opened 2020-02-21
Added a new compact result display mode

### LINKED ISSUE #935: Display more calculations on screen

0b101:
This adds a compact result mode which displays the input and output of the calculation on the same line if they can both fit. It also allows for more calculations to be displayed at once.

Before: 
E4

After:
E1E10E11

It doesn't mess with the sin/cos/tan display
E5

This is based on #954 and https://github.com/Omega-Numworks/Omega/pull/238
This closes #935


apps/calculation/calculation.cpp
```diff
@@ -182,23 +192,39 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
...
     KDCoordinate approximateOutputHeight = approximateLayout.layoutSize().height();
+    KDCoordinate approximateOutputWidth = approximateLayout.layoutSize().width();
+    singleLine = approximateOutputWidth + inputWidth < maxWidth - 40;
     if (displayOutput(context) == DisplayOutput::ApproximateOnly || (!expanded && displayOutput(context) == DisplayOutput::ExactAndApproximateToggle)) {
-      if (allExpressionsInline) {
-        KDCoordinate approximateOutputBaseline = approximateLayout.baseline();
-        result = maxCoordinate(inputBaseline, approximateOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, approximateOutputHeight-approximateOutputBaseline);
+      
+      if (singleLine && Poincare::Preferences::sharedPreferences()->resultDisplay() == Poincare::Preferences::ResultDisplay::Compact && !allExpressionsInline) {
+          result = (inputHeight >= approximateOutputHeight) ? inputHeight + singleMargin : approximateOutputHeight + singleMargin;
```
0b101:
```suggestion
        result = (inputHeight >= approximateOutputHeight) ? inputHeight + singleMargin : approximateOutputHeight + singleMargin;
```

Ecco:
Thank you @0b101 ! Definitely something we'll want to merge, thanks!

I haven't looked at the code just yet, but we'll need to make sure this works perfectly well before merging because we'll most likely make this the default (and only) behavior. 🎉

As a result, I think you can remove all the settings-related code already.

That being said, it seems like there is a baseline issues in the `11^8` example.

0b101:
@Ecco I think that it's trying to center the answer for `11^8` with `11^8`. I'll take a look look into moving it down to the same level.

0b101:
The baseline bug and a bug where results sometimes show up as "t" is now fixed!
SCF

apps/calculation/calculation.cpp
```diff
@@ -156,11 +160,18 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
 
   if (displayOutput(context) == DisplayOutput::ExactOnly) {
     KDCoordinate exactOutputHeight = exactLayout.layoutSize().height();
-    if (allExpressionsInline) {
+    KDCoordinate exactOutputWidth = exactLayout.layoutSize().width();
+    singleLine = exactOutputWidth + inputWidth < maxWidth - 40;
+    if (singleLine && !allExpressionsInline) {
       KDCoordinate exactOutputBaseline = exactLayout.baseline();
-      result = maxCoordinate(inputBaseline, exactOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, exactOutputHeight-exactOutputBaseline);
+      result = maxCoordinate(inputBaseline, exactOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, exactOutputHeight-exactOutputBaseline) + singleMargin;
```
0b101:
```suggestion
      result = maxCoordinate(inputBaseline, exactOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, exactOutputHeight-exactOutputBaseline);
```

apps/calculation/history_view_cell.cpp
```diff
@@ -185,9 +185,11 @@ void HistoryViewCell::layoutSubviews(bool force) {
     inputSize.height()),
   force);
   KDSize outputSize = m_scrollableOutputView.minimalSizeForOptimalDisplay();
+  int singleLine = outputSize.width() + inputSize.width() < bounds().width() - 6;
+  int outputHeight = (singleLine) ? (maxCoordinate(0, inputSize.height() - outputSize.height()) / 2) + maxCoordinate(0, (inputSize.height() - outputSize.height()) / 2) : inputSize.height();
```
0b101:
```suggestion
  int outputHeight = (singleLine) ? (maxCoordinate(0, inputSize.height() - outputSize.height()) / 2) + maxCoordinate(0, (inputSize.height() - outputSize.height()) / 2) + 1 : inputSize.height();
```

0b101:
After extensive testing, it appears that everything is now working as expected. If you see anything that looks “off”, try to first test if you get the same thing in the HP Prime emulator as that’s what the compact result display is based off and then feel free to comment.

EbbDrop:
Using str-> to save a function seems to be broken. The underside is cutoff.
Screenshot_20200308-145610

0b101:
@EbbDrop Oh! Thank you for pointing this out! I must have forgotten to add the margin. I’ll do that when I have the chance.

apps/calculation/calculation.cpp
```diff
@@ -156,11 +160,18 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
 
   if (displayOutput(context) == DisplayOutput::ExactOnly) {
     KDCoordinate exactOutputHeight = exactLayout.layoutSize().height();
-    if (allExpressionsInline) {
+    KDCoordinate exactOutputWidth = exactLayout.layoutSize().width();
+    singleLine = exactOutputWidth + inputWidth < maxWidth - 40;
+    if (singleLine && !allExpressionsInline) {
       KDCoordinate exactOutputBaseline = exactLayout.baseline();
       result = maxCoordinate(inputBaseline, exactOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, exactOutputHeight-exactOutputBaseline);
```
0b101:
```suggestion
      result = maxCoordinate(inputBaseline, exactOutputBaseline) + maxCoordinate(inputHeight - inputBaseline, exactOutputHeight-exactOutputBaseline) + singleMargin;
```

artaxxx:
Thanks @0b101 , we're going to merge this PR!
Before that, I think there's a problem with the margin regarding one of the new features (additional results). Just type 1+1+1+1... (13 times) and go up in the history to see the bug. Please, can you have a look at this?

0b101:
Thanks for pointing that out @artaxxx. It is now fixed. I guess I forgot to add on the ellipsis' width when it was introduced.

Ecco:
Size analysis failed because this branch hasn't been rebased on top of a recent master

0b101:
> Size analysis failed because this branch hasn't been rebased on top of a recent master

@Ecco It's now rebased and it can be merged cleanly.

apps/calculation/calculation.cpp
```diff
@@ -156,11 +160,18 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
 
   if (displayOutput(context) == DisplayOutput::ExactOnly) {
     KDCoordinate exactOutputHeight = exactLayout.layoutSize().height();
-    if (allExpressionsInline) {
+    KDCoordinate exactOutputWidth = exactLayout.layoutSize().width();
+    singleLine = exactOutputWidth + inputWidth < maxWidth - 40 - 37;
```
EmilieNumworks, line 164: Could you express these magic numbers in terms of other object widths? I haven't investigated which ones but maybe: Metric::EllipsisCellWidth, Metric::CommonSmallMargin?

apps/calculation/calculation.cpp
```diff
@@ -182,23 +193,38 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
     }
 
     KDCoordinate approximateOutputHeight = approximateLayout.layoutSize().height();
+    KDCoordinate approximateOutputWidth = approximateLayout.layoutSize().width();
+    singleLine = approximateOutputWidth + inputWidth < maxWidth - 40 - 37;
```
EmilieNumworks, line 197: Same comment

apps/calculation/calculation.cpp
```diff
@@ -182,23 +193,38 @@ KDCoordinate Calculation::height(Context * context, bool expanded, bool allExpre
...
+          result = inputHeight + approximateOutputHeight + doubleMargin;
+        }
       }
     } else {
       assert(displayOutput(context) == DisplayOutput::ExactAndApproximate || (displayOutput(context) == DisplayOutput::ExactAndApproximateToggle && expanded));
       KDCoordinate exactOutputHeight = exactLayout.layoutSize().height();
       KDCoordinate exactOutputBaseline = exactLayout.baseline();
+      KDCoordinate exactOutputWidth = exactLayout.layoutSize().width();
+      KDCoordinate approximateOutputWidth = approximateLayout.layoutSize().width();
+      singleLine = exactOutputWidth + approximateOutputWidth + inputWidth < maxWidth - 70 - 37;
```
EmilieNumworks, line 217: Same comment

apps/calculation/calculation.h
```diff
@@ -95,6 +95,7 @@ friend CalculationStore;
   // Additional Information
   AdditionalInformationType additionalInformationType(Poincare::Context * context);
 private:
+  static constexpr int maxWidth = 314;
```
EmilieNumworks, line 98: This should be something like Ion::Display::Width - something?

apps/calculation/history_view_cell.cpp
```diff
@@ -186,9 +186,11 @@ void HistoryViewCell::layoutSubviews(bool force) {
     inputSize.height()),
   force);
   KDSize outputSize = m_scrollableOutputView.minimalSizeForOptimalDisplay();
+  int singleLine = outputSize.width() + inputSize.width() < bounds().width() - 6 - Metric::EllipsisCellWidth;
```
EmilieNumworks, line 189: Same comment about magic numbers

apps/calculation/calculation.cpp
0b101: I tried to do it with Metric::EllipsisCellWidth and the text would go invisible/ be hidden behind the ellipsis whenever it was selected.

0b101: I changed "-37" to Metric::EllipsisCellWidth in my latest commit. the "-40" or "-70" is still required though

apps/calculation/calculation.h
0b101: I changed "314" to "Ion::Display::Width - 6" in my latest commit

Ecco:
screenshot (7)

There's a small centering glitch here. The `1.2` on the left should be vertically aligned with the one on the right.

0b101:
@Ecco This is actually intentional behavior so the user can see the result when there is a tall input. It also matches the way that the HP Prime displays the input and output
PRM

Ecco:
Hmm, I still think the baselines should be aligned. (Very) tall inputs might be a problem on their own (we'd most likely need to have a better scrolling mechanism), but I really think the "simple" case should be aligned.

0b101:
Ok, I’ll see what I can do.

Ecco:
> Ok, I’ll see what I can do.

Awesome. We definitely can help if needed!

0b101:
@Ecco How does this look? I still have to do a little more testing, but I think it looks good.
Bottom

adriweb:
Heh, my order of alignment preference would be center, top, bottom.
Also, what would happen if: the input is short (in both H and W) but the answer is very tall (but still has quite a short width). Would the input be placed out of the view? I guess it should somehow be always visible?

0b101:
> Also, what would happen if: the input is short (in both H and W) but the answer is very tall (but still has quite a short width). Would the input be placed out of the view? I guess it should somehow be always visible?

The input is now bottom aligned with the answer so it should still be entirely visible as long as it's not so tall that the top goes out of view. If you have a input that you want me to enter, I'll be glad to send you a screenshot?

Ecco:
> @Ecco How does this look?
> Bottom

Well, it looks… bottom aligned, I guess 😄

Ecco:
> What would happen if: the input is short (in both H and W) but the answer is very tall (but still has quite a short width). Would the input be placed out of the view? I guess it should somehow be always visible?

I really think we shouldn't bother with scrolling issues on this PR since we *already* don't handle (abnormally) high results very well anyway.

Ecco:
Here are the issues we'd like to fix before merging this PR:
- [ ] Make sure inputs and outputs are baseline-centered
- [ ] Increase the vertical margins around results (6px -> 9px) (See screenshot)
- [ ] Properly handle selection to make sure it's "geographic" (currently in single-line mode, pressing "right" from an input goes straight to the "…" icon)

@0b101 : would you mind if we modify your PR or would you rather take care of this by yourself?

0b101:
> would you mind if we modify your PR or would you rather take care of this by yourself?

@Ecco feel free to modify it to suit your needs

Ecco:
After giving it some thought, a 12px margin is even better. So instead of using CommonSmallMargin=5 we should use a value of 12. Maybe we would like to increase CommonLargeMargin to 12 for everywhere else?

0b101:
Will both the commonSmallMargin and the commonLargeMargin = 12? or will the smallMargin use a different value?

0b101:
After some investigation, it looks like 9px would be better than 12. With 12px, only 4 small calculations or 2 big calculations can be shown. That’s about the same as the old, non-compact display, where there are only 3 calculations shown. 9px, on the other hand, can fit 5 small calculations while still not looking too crowded. 

If you want to see how 12px displays less large calculations than the current firmware, type in `ln(10)` a couple times.

Ecco:
We did several tests before commenting, including with 10 and 12px. It's true that more results are visible if the margin is smaller, but it's also harder to tell results apart.

Obviously it's a matter of personal taste, but we think 12px is better for most people.

0b101:
That makes sense. I just committed a change so the margin change won't break anything.

LeaNumworks:
Hi @0b101,
Thanks for this PR! I rebased it and fixed some issues in #1569 (I was not able not push to your repo), so I am closing this PR.

PR #1346, opened 2020-02-22
[ion] Fix error check in the USB stack

### LINKED ISSUE #1335: ion/src/device/shared/usb/stack/endpoint0.cpp:146: comparison tautologically false
In ion/src/device/shared/usb/stack/endpoint0.cpp, on lines 146 and 154, the code does not have the intended effect. The return type of the receiveSomeData function is unsigned, so it will never be negative, and the comparison is always false.

BTW, you people should compile with -Wextra, to catch bugs like this one in the future.

nsajko:
The return type of the function needs to be wider so the -1 error value
could fit.

Fixes #1335


LeaNumworks:
Thanks for this PR!

Issue #1290, opened 2020-01-31
Add script's size in Script Option


A2drien:
#### Describe the solution you'd like
Add python script's size.

#### Describe alternatives you've considered
Add python script's size in Script Option

artaxxx:
Finally, I'm closing this issue. See #1487

PR #1347, opened 2020-02-22
Added a script size cell in script parameter

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1290

## The result (in red and with another button)
image
By hitting the ok button you get that :
image

## Todo :
- [x] Add the translations


RedGl0w:
Woops missclick !

Issue #1023, opened 2019-06-20
Python script list display: more scripts and size


XanderLeaDaren:
#### Problem you'd like to fix
The actual display (list) of Python scripts lacks the size of these scripts and can only display three scripts at a time.

#### Screenshots
Left: actual / Right: mock-up proposal
![python-scripts-list](https://user-images.githubusercontent.com/28948599/59851522-23a14280-936d-11e9-849e-c8e06d7b0d36.jpg)

#### Describe the solution you'd like
1. Height of every line: 50 px instead of actual 100 px
2. End of line: show every script size, in octets

#### Describe alternatives you've considered
No alternative for now

0b101:
The 50px line heights look a little squished. A 75px line height might look a little better while still allowing more scripts to be displayed.

boricj:
You also need to display the available/total storage capacity, otherwise the script sizes by themselves are not really a useful information for the end-user.

XanderLeaDaren:
Both of your ideas are good!

Capacity seems related to #623 “No size or memory limit mentioned”

RemyDCF:
The height is currently defined by Metric::StoreRowHeight, which is currently system-wide, so if we want to change the cell's height, we can either change the Metric::StoreRowHeight, or we can create a new variable.
And, in my opinion, setting Metric::StoreRowHeight to 37 is quite a good fit.

<img width="506" alt="Capture d’écran 2019-06-20 à 16 25 41" src="https://user-images.githubusercontent.com/11600393/59856995-4b49d800-9378-11e9-9cd2-55d8e3ab0f25.png">

0b101:
> The height is currently defined by Metric::StoreRowHeight, which is currently system-wide

Does this change anything other than the Python script list?

RemyDCF:
> > The height is currently defined by Metric::StoreRowHeight, which is currently system-wide
> 
> Does this change anything other than the Python script list?

Functions, equations and sequences list, at least. But the height is adapted if the cell need a bigger height (for fractions, square roots, ...)

XanderLeaDaren:
With these new settings, I remark the bottom buttons are not consistent in height: “Solve the system“ and “Python shell” appear with a bigger font. For the latter, using the smaller font (as for “Plot graph” and “Display values”) is another way of getting one more script to appear in the list. What do you think?

artaxxx:
Let me close this issue since we decided not to display script size in the app.

artaxxx:
[...]

PR #1349, opened 2020-02-23
[apps/code] Script size 

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1290

## The result (in red and with another button)
image
By hitting the ok button you get that :
image

## Todo :
- [x] Add the translations

Note : this is a reopen of https://github.com/numworks/epsilon/pull/1347


artaxxx:
Thanks @RedGl0w! I'm still wondering if it's the best place to display the size of the script but I flashed a device with your PR and I noticed some little things:
* as the size is an information, it should be in grey and right-aligned (see Software version in Settings)
* I'm not sure the text will fit if you want to display "32768o/32768o" in the different languages
* some unexpected characters are displayed if I press OK several times
I'm continuing to think about whether it's appropriate to display this information in the options menu.

RedGl0w:
> I'm continuing to think about whether it's appropriate to display this information in the options menu.

When there was only 16Ko for Ion::Storage, it was very usefull... I think it's still usefull, and should be displayed here, or like https://github.com/numworks/epsilon/issues/1023 says, next to the file name... What do you think is best ?

RedGl0w:
> I'm continuing to think about whether it's appropriate to display this information in the options menu.

If it shouldn't be displayed in the options menu, where would you like to see it ?

RedGl0w:
Rebased : https://github.com/numworks/epsilon/pull/1487

Issue #1173, opened 2019-10-31
Warn for illegals character in i18n files


RedGl0w:
#### Problem you'd like to fix
I would like to stop the compile when you insert illegal characters in i18n files (like ⁻¹) because it doesn't crash, but all the message will be buggy...

EmilieNumworks:
Fix by #1459

PR #1359, opened 2020-02-25
[I18n] Warn for illegal characters

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1173
## An example of logs :
```
I18N    apps/i18n.cpp
👍 (0x128077) is not a valid character
```
## How it works :
It just parse the code_points.h file, and check for each character if it is inside this file


Ecco:
That's interesting, thank! Out of curiosity, in which scenario would this be useful in practice?

RedGl0w:
> That's interesting, thank! Out of curiosity, in which scenario would this be useful in practice?

On omega we're doing a support for Hungarian, and we got wrong character, but we discovered only when I created and tested this PR

RedGl0w:
> > That's interesting, thank! Out of curiosity, in which scenario would this be useful in practice?
> 
> On omega we're doing a support for Hungarian, and we got wrong character, but we discovered only when I created and tested this PR

Also, I just remembered when we added some physic constants, I tried the ^-1 character, and realized that there were no errors given, and I found that it was strange...

RedGl0w:
Any news @Ecco ?

kandinsky/Makefile
```diff
@@ -25,6 +25,10 @@ tests_src += $(addprefix kandinsky/test/,\
   rect.cpp\
 )
 
+code_points += $(addprefix kandinsky/fonts/, \
+  code_points.h \
```
EmilieNumworks, line 29: `code_points = kandinsky/fonts/code_points.h`

EmilieNumworks:
Good idea @RedGl0w. I'd like to merge this, could you please resolve the conflicts with the master branch? Could you also avoid the "Merge" commit. When possible, we would rather keep a "clean" memory.

RedGl0w:
I'll recreate my PR, because I had to delete my repo, and so I can't commit anymore :/

Issue #1297, opened 2020-02-08
Be able to make color codes with integers


A2drien:
#### Problem you'd like to fix
Since version 11, a color code consists of a tuple instead of an integer.
The problem is that now many scripts no longer work properly.

#### Describe the solution you'd like
Make sure you can choose between a color code with tuple / integer.

RedGl0w:
![image](https://user-images.githubusercontent.com/43498612/75421014-9ff71700-5939-11ea-8574-6e4ad5566e24.png)
Somethink like that ?

PR #1364, opened 2020-02-27
Kandinsky and turtle support int as colors

RedGl0w:
This patch, I think, https://github.com/numworks/epsilon/issues/1297. You can now add an int as color in `color()` for turtle and `fill_rect()`,`set_pixel()`, like `0xFF00FF`.


Ecco:
Hi @RedGl0w ! Thanks a lot for this PR! This is definitely a great addition, and we'd like to merge something very similar albeit subtly different.

I wrote an extensive definition on what we'd want to merge in #1533 .

As a result I'm going to close this PR, but feel free to:
 * Discuss the specification in #1533 
 * Create a new PR that matches #1533 

Thanks!!

PR #1367, opened 2020-02-27
Patched some bad units

RedGl0w:
Yes a month is 30 weeks :trollface: https://github.com/numworks/epsilon/blob/f12c53b3f2c76f0970f32ec5ca006a3f495c3077/poincare/include/poincare/unit.h#L238
Reported by https://github.com/A2drien


poincare/include/poincare/unit.h
```diff
@@ -226,19 +226,19 @@ class Unit final : public Expression {
...
-        Representative("week", "7*24*60*60*_s",
+        Representative("week", "7*_day",
             Representative::Prefixable::No,
             NoPrefix),
-        Representative("month", "30*7*24*60*60*_s",
+        Representative("month", "(365.25/12)*_day", // A month is a twelveth of a year
             Representative::Prefixable::No,
             NoPrefix),
-        Representative("year", "365.25*24*60*60*_s",
+        Representative("year", "365.25*_day",
```
RedGl0w:
```suggestion
        Representative("year", "365*_day",
```

poincare/include/poincare/unit.h
```diff
@@ -226,19 +226,19 @@ class Unit final : public Expression {
...
-        Representative("day", "24*60*60*_s",
+        Representative("day", "24*_h",
             Representative::Prefixable::No,
             NoPrefix),
-        Representative("week", "7*24*60*60*_s",
+        Representative("week", "7*_day",
             Representative::Prefixable::No,
             NoPrefix),
-        Representative("month", "30*7*24*60*60*_s",
+        Representative("month", "(365.25/12)*_day", // A month is a twelveth of a year
```
RedGl0w:
```suggestion
        Representative("month", "(360/12)*_day", // A month is a twelveth of a year
```
Or maybe something like
```
 Representative("month", "30*_day",
```

RubenNumworks:
Thank you @A2drien and @RedGl0w.

1. The definition of the month unit must obviously be corrected. The other changes are of a different nature.
2. On the one hand, changing the definition of the `_h` unit from `60*60_s` to `60_min` makes indeed the code more readable and seems harmful. However it slows down to handling of units.
3. On the other hand, you raise the question of the definition of the `_month` and `_year` units.

**Could you please split your pull request into 3 commits, one for each item of the above list?** That way
1. would certainly be merged
2. probably not (unless the handling of units is optimized)
3. needs to be discussed, see below.

There is no general standard definition of the duration of a month and of the year, certainly not in the _International System of Units (SI)_. Those units are very convenient for human understanding (for instance, `5.03_year` is more understandable by humans than `158947641_s`). On the contrary, an experimental physicist (not the intended public for our calculator) wanting to measure some duration with high precision would never give their results in years or months. Hence I guess that there is no significant difference in defining the year unit as 365 or 365,25 days and similarly the month unit as 30 days or one twelveth of a year. If a precise result is required, just use a standard unit. We went for `1 month = 30 days` and `1 year = 365.25 days`. Do you think that some other definition would be better?

RedGl0w:
> **Could you please split your pull request into 3 commits, one for each item of the above list?** That way

That's done

> If a precise result is required, just use a standard unit. We went for `1 month = 30 days` and `1 year = 365.25 days`. Do you think that some other definition would be better?

I don't know...
I think that `1 month = 30 days` is good, but not enough... It wouldn't be very user friendly, but I find that `1 month = 365.25 days / 12` is a way better definition

RubenNumworks:
> I think that `1 month = 30 days` is good, but not enough... It wouldn't be very user friendly, but I find that `1 month = 365.25 days / 12` is a way better definition

Fair enough.
Would you mind making a clean pull request only with
1. your commit (f37d2e7) where you fix the definition of the month unit,
2. a second commit where you change the definition from `30*24*60*60*_s` to `30*24*60*60/12*_s` ?
Or I may do it.

As for the last part, we should wait for an optimization of units before considering code readability.

RedGl0w:
> Would you mind making a clean pull request only with
> 
>     1. your commit ([f37d2e7](https://github.com/numworks/epsilon/commit/f37d2e7deb5504195addaef1e60a7d2a3bd32d76)) where you fix the definition of the month unit,
> 
>     2. a second commit where you change the definition from `30*24*60*60*_s` to `30*24*60*60/12*_s` ?

Done with #1382 !

PR #1369, opened 2020-02-28
Remove useless uses of const for return types

nsajko:
An integer or float value can not be assigned to, so there is no point
in qualifying it as const.


EmilieNumworks:
You're right, thanks @nsajko!

Issue #1350, opened 2020-02-24
Plotting function: incorrect graph


JuniorDelaure:
Version 12.4.0 
This function's graph is incorrect 

x^2.0000001

![IMG_20200224_094912](https://user-images.githubusercontent.com/56046513/75138799-68932b00-56eb-11ea-92b6-d87aa9e7f757.jpg)

![IMG_20200224_094852](https://user-images.githubusercontent.com/56046513/75138788-6204b380-56eb-11ea-9e32-c06d10c45eff.jpg)

nsajko:
This is kind of maybe similar to how second-derivatives (of any function, I think) get graphed. That is, the *steps* are too large in that case, too.

EmilieNumworks:
Good catch @JuniorDelaure, thanks! There are indeed two issues here: first, the function should not be defined for x < 0 as its imaginary part is not null and, second, there is a major precision loss on the function plotting...

1. When computing power of complexes, we sometimes neglect the imaginary part or the real part to avoid weird result as `exp(iπ) = -1+6E-17*i`. However, the decision to neglect is based on an absolute comparison when it should depends on the values of the complexes (Cf comment and code in `poincare/src/power.cpp` on method  `PowerNode::compute`). For instance, (-10)^0.00000001 ~ 1 when it should be ~1+3.141593ᴇ-8*i

2. When simplifying x^(p/q), with p, q integers, before an approximation, we turn it into (nth_root(x,q))^p. This was done in order to be able to choose another root than the principal root if the complex mode is real. For instance, (-8)^(1/3) returns -2 in real mode and 1+sqrt(3)*i in cartesian mode. However, in our case, f(x) = x^(2.0000001) = x^(20000001/10000000) = nth_root(x, 10000000)^20000001. This has the side effect of raising the precision error in nth_root(x, 10000000) to the power 20000001... This explains the weird look of the curve.

I've made 2 patches to fix these errors, I'll make a PR soon and refer it here.

EmilieNumworks:
PR #1360 should resolve this issue.

artaxxx:
![screenshot (52)](https://user-images.githubusercontent.com/20040032/78790856-680ad700-79af-11ea-8b0f-35224a667f30.png)
It's better now, but it's not perfect! The function isn't defined for x<0 but you can see a part of the graph...

Issue #1222, opened 2019-11-27
[Bug] Failed to graph second derivative of 3x/2e^x+e^-x


TheWanderingCoel:
#### Describe the bug
A clear and concise description of what the bug is. Please describe a **single** bug per issue. Feel free to create multiple issues though!

I used my TI-83 and it's no problem.  
When I use numworks calculator, It stucked at ``plot graph`` button clicked. And after a few minutes, a strange graph was produced.

#### Screenshots

![WX20191127-135149](https://user-images.githubusercontent.com/39617891/69697320-160e9680-111d-11ea-8300-997f52902e59.png)

![WX20191127-135405](https://user-images.githubusercontent.com/39617891/69697435-6c7bd500-111d-11ea-8b46-46996b87c78a.png)

#### To Reproduce
1. Go to the 'Functions' app
2. Type '3x/2e^x+e^-x'
3. Click 'plot graph'

#### Expected behavior

Graph correctly

![WX20191127-135618](https://user-images.githubusercontent.com/39617891/69697556-bbc20580-111d-11ea-894c-25547a3a2057.png)

#### Environment
 - Epsilon version (Settings > About > Software version).
 - The platform(s) on which the problem happens: online simulator, actual device, etc...

Both on online and actual device

nsajko:
I tried this on my device (which is on 984d1ff), and this turns out to be not just an issue of the graph being bad, I had to reset my Numworks because it was frozen, neither the back button nor the power button worked while the graph was being "drawn".

TheWanderingCoel:
> I tried this on my device (which is on [984d1ff](https://github.com/numworks/epsilon/commit/984d1ff924e85e111439d1220ed7da4ed192d6ed)), and this turns out to be not just an issue of the graph being bad, I had to reset my Numworks because it was frozen, neither the back button nor the power button worked while the graph was being "drawn".

Same here, you had to reset your calculator otherwise you need wait a long time...

suazhero:
Also occurs with (x^3)/6. Supposed to plot y=x when derivated twice. But plots the following instead
But is it supposed to be supported at all? 

![image](https://user-images.githubusercontent.com/55035114/70997056-6ea9d180-20d4-11ea-9416-82a34ac97eb3.png)

TheWanderingCoel:
It should be supported. But I don't if there was a bug causing graph failure.

suazhero:
There's clearly something weird as soon as function is defined from two times derivatives.
[...]

PR #1360, opened 2020-02-26
Poincare fix precision errors

EmilieNumworks:
- Fix wrong power approximation: (-10)^0.00000001 ~ 1 instead of ~1+3.141593ᴇ-8*i
- Fix precision loss in power approximation
See issue #1350 for more details

PR #1376, opened 2020-03-02
[poincare] Prevent unnecessary errors of mathematical functions

### LINKED ISSUE #1373: Complex atanh loses imaginary part of result
#### Describe the bug
See the screenshot, the result is shown as a real number (note that the number does correctly correspond to both real(atanh(given input)) and to atanh(real(given input))). Considering Calculation uses double precision floating point/binary64 (which can represent numbers like 1e-320) and imag(correct result) is approximately 5.0e-16, this is a significant error.

Comparisons with C++ standard library (GNU) and Go(lang) standard library, which are to be expected to give similar results as Numworks Epsilon considering all of them use binary64, are given below. Five input values were chosen for this test, all of which produce OK results with both Go and C++, while Epsilon fails miserably.

C++ source:
```
#include <iostream>
#include <complex>
 
int
main() {
#define iR 0.99999999999
#define oR 13.010791601744723916808
	std::complex<double> i0(iR, 1.0e-26), i1(iR, 1.0e-60), i2(iR, 1.0e-150), i3(iR, 1.0e-250), i4(iR, 1.0e-300),
		o0(oR, 5.0000000000250000000001e-16),
		o1(oR, 5.0000000000250000000001e-50),
		o2(oR, 5.0000000000250000000001e-140),
		o3(oR, 5.0000000000250000000001e-240),
		o4(oR, 5.0000000000250000000001e-290);

	std::cout <<
		std::atanh(i0) << '\n' <<
		std::atanh(i0) - o0 << '\n' << '\n' <<
		std::atanh(i1) << '\n' <<
		std::atanh(i1) - o1 << '\n' << '\n' <<
		std::atanh(i2) << '\n' <<
		std::atanh(i2) - o2 << '\n' << '\n' <<
		std::atanh(i3) << '\n' <<
		std::atanh(i3) - o3 << '\n' << '\n' <<
		std::atanh(i4) << '\n' <<
		std::atanh(i4) - o4 << '\n';
}
```

C++ std:atanh results and errors (same with `-O0`, `-O3`, or without optimization options):
```
(13.0108,5e-16)
(-4.13702e-08,-4.13702e-23)

(13.0108,5e-50)
(-4.13702e-08,-4.13702e-57)

(13.0108,5e-140)
(-4.13702e-08,-4.13702e-147)

(13.0108,5e-240)
(-4.13702e-08,-4.13702e-247)

(13.0108,5e-290)
(-4.13702e-08,-4.13702e-297)
```

Go playground link for Go source and output: https://play.golang.org/p/vvQd4NQoW3t

Go cmplx.Atanh results and errors:
```
Result: (13.01079156037454+4.99999958629818e-16i)
Error: (-4.137018372318835e-08-4.137268206283146e-23i)

Result: (13.01079156037454+4.999999586298179e-50i)
Error: (-4.137018372318835e-08-4.137268205797551e-57i)

Result: (13.01079156037454+4.99999958629818e-140i)
Error: (-4.137018372318835e-08-4.137268199607741e-147i)

Result: (13.01079156037454+4.999999586298179e-240i)
Error: (-4.137018372318835e-08-4.137268211085086e-247i)

Result: (13.01079156037454+4.9999995862981794e-290i)
Error: (-4.137018372318835e-08-4.137268206639525e-297i)
```

#### Screenshots
atanh_input

#### To Reproduce
1. Go to the Calculation app
2. Type as in screenshot, or choose another of the selected five inputs, or a similar input

#### Expected behavior
A result in line with correctness achieved by other functions that use binary64 from common libraries.

#### Environment
 - Epsilon versions 12.3.0, 13.0.0
 - The platform(s) on which the problem happens: online simulator, actual device

nsajko:
This removes the two functions Trigonometry::RoundToMeaningfulDigits and
ApproximationHelper::TruncateRealOrImaginaryPartAccordingToArgument (and
their uses) which are used for rounding the results of almost all
mathematical functions exposed to the user on the calculator. Also
removed is a similar procedure used inline after the computation of
derivatives.

The intent of the removed code was to round the results of mathematical
functions with the supposed accuracy of the underlying liba (OpenBSD
libm) and libaxx (libcxx) functions in mind, so no significant error
would be introduced.

The functions have their implementation and interface issues: they can
create a much larger error of the result of the mathematical function
than intended and fixing them to work as intended would necessitate a
lot of work and increasing code size, as each mathematical function
would need its own "rounding function" to prevent errors greater than
intended.

However, even if the removed functions were made to work as intended at
whatever cost, their use in Poincare in the current manner could still
produce horrible results. This is because even insignificant errors can,
with propagating the rounded outputs of functions to other functions
which again round their outputs, ruin the total result at the end of the
calculation (because of a greatly magnified error).

Also note that the problem the "rounding functions" were intended to
solve can be solved in a simpler manner by giving the user the (perhaps
default) option of a fixed point fractional number display (like the %f
conversion specifier in C printf), coupled with more usage of tables of
notable values of mathematical functions.

Fixes #1373

Updates #1350 (The left part of the plot with the imaginary part is
gone, but the right part's plot still has the steps/staircase. The
values shown for each x are accurate and do not correspond to the
staircase, though.)

Updates #1222 (The plot is a lot nicer, but still has some gaps.)

### DIFF
--- a/poincare/include/poincare/approximation_helper.h
+++ b/poincare/include/poincare/approximation_helper.h
@@ -10,7 +10,6 @@ namespace Poincare {
 
 namespace ApproximationHelper {
   template <typename T> int PositiveIntegerApproximationIfPossible(const ExpressionNode * expression, bool * isUndefined, Context * context, Preferences::ComplexFormat complexFormat, Preferences::AngleUnit angleUnit);
-  template <typename T> std::complex<T> TruncateRealOrImaginaryPartAccordingToArgument(std::complex<T> c);
 
   template <typename T> using ComplexCompute = Complex<T>(*)(const std::complex<T>, Preferences::ComplexFormat complexFormat, Preferences::AngleUnit angleUnit);
   template<typename T> Evaluation<T> Map(const ExpressionNode * expression, Context * context, Preferences::ComplexFormat complexFormat, Preferences::AngleUnit angleUnit, ComplexCompute<T> compute);
--- a/poincare/include/poincare/trigonometry.h
+++ b/poincare/include/poincare/trigonometry.h
@@ -23,9 +23,6 @@ public:
   static Expression table(const Expression e, ExpressionNode::Type type,  ExpressionNode::ReductionContext reductionContext); // , Function f, bool inverse
   template <typename T> static std::complex<T> ConvertToRadian(const std::complex<T> c, Preferences::AngleUnit angleUnit);
   template <typename T> static std::complex<T> ConvertRadianToAngleUnit(const std::complex<T> c, Preferences::AngleUnit angleUnit);
-  template <typename T> static std::complex<T> RoundToMeaningfulDigits(const std::complex<T> result, const std::complex<T> input);
-private:
-  template <typename T> static T RoundToMeaningfulDigits(T result, T input);
 };
 
 }
--- a/poincare/src/approximation_helper.cpp
+++ b/poincare/src/approximation_helper.cpp
@@ -27,18 +27,6 @@ template <typename T> int ApproximationHelper::PositiveIntegerApproximationIfPos
   return absInt((int)scalar);
 }
 
-template <typename T> std::complex<T> ApproximationHelper::TruncateRealOrImaginaryPartAccordingToArgument(std::complex<T> c) {
-  T arg = std::arg(c);
-  T precision = 10*Expression::Epsilon<T>();
-  if (absMod<T>(arg, (T)M_PI) <= precision) {
-    c.imag(0);
-  }
-  if (absMod<T>(arg-(T)M_PI/2.0, (T)M_PI) <= precision) {
-    c.real(0);
-  }
-  return c;
-}
-
 template<typename T> Evaluation<T> ApproximationHelper::Map(const ExpressionNode * expression, Context * context, Preferences::ComplexFormat complexFormat, Preferences::AngleUnit angleUnit, ComplexCompute<T> compute) {
   assert(expression->numberOfChildren() == 1);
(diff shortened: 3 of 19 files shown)


EmilieNumworks:
Hi @nsajko,
As you reported in the issue #1373, these two functions - `Trigonometry::RoundToMeaningfulDigits` and
`ApproximationHelper::TruncateRealOrImaginaryPartAccordingToArgument` - lead to quite wrong approximations. However, removing  them utterly would imply that e^(iπ) != -1 and sin(π) != 0 when approximating (see the broken unary tests). And these computations are quite common in high school...
The key point with these functions is when to decide wether the real (or imaginary) part should be neglected. In `ApproximationHelper::TruncateRealOrImaginaryPartAccordingToArgument` current implementation, the comparison is absolute! It should be done relatively to the input.
The PR #1360 fixes this. I also added your tests on `atanh` in our unary tests.

EmilieNumworks:
This choice is of course a trade-off. Removing any kind of corrections on libc functions (cos, sin, tan...) would lead to this kind of issues when drawing `cos(x)`:
Capture d’écran 2020-03-06 à 10 05 52

However, keeping them, even with their new implementation would introduce some errors like (when plotting `x^2.00001`):
Capture d’écran 2020-03-06 à 10 11 11
(the plot should be undefined for x < 0)

I choose to keep the latter errors for several reasons:
- cos/sin/tan plot are more common than x^2.00001 in high-school
- Anyway, `x^2.00000001` will be displayed:
Capture d’écran 2020-03-06 à 10 19 11
Because the IEEE754 float representation of 2.00000001 is 2.


If you have other examples of errors that PR #1360 would let, please let me know. I am still a bit unsatisfied with the current solution!

Issue #624, opened 2018-08-13
Matrix containing matrix doesn't work


menthalo37:
![capture2matrices](https://user-images.githubusercontent.com/42341413/44034254-f27b02ae-9f0c-11e8-80f3-ab7dd6578165.JPG)

In the following test, M1 was previously initialized by [[1,2],[3,4]] -> M1.
Sub matrix doesn't seem to work properly.

PR #1380, opened 2020-03-02
[poincare/matrix] Forbid nested matrices

LeaNumworks:
Scenario: in calculation,
1/matrix(matrix(matrix(matrix(... matrix(1) ... )))) gave kind of a
weird result. These kind of computations also broke the fuzzer.


RubenNumworks:
Fixes #624

Issue #931, opened 2019-03-23
Graph color option. (Colorblind support)


Pixel-Null:
Would it be possible to add a color option in the function app for each different function?
So you'd be able to select colors for a function.
with a color menu added to the function options or the use of arrows in a specific key combination to cycle through the available colors?

I'm personally colorblind and i have a hard time seeing the difference between them, and having to add and remove functions until i get a combination of colors i can clearly see the difference between can be quite the hassle.

Thanks in advance!

boricj:
Do you have any other issues with other screens? I've raised the question of accessibility before (#597), but I think this is the first time we actually have a user that is affected by this on hand.

PnDevs:
I do yes, on every screen (red and green colorblind), some of them have colorblind filters built in but honestly it's just an issue in graphs at the moment so i wouldn't bother making a filter for the entire calculator.

DSchndr:
Maybe making an colorblind palette in escher, which can be enabled from the settings menu could be the solution for this problem.
Because then you only have to change a couple of values that are used globally.

When i have time i'll experiment a little bit with the idea and ask one from my class (who is r/g colorblind) if it is better.

boricj:
The NumWorks team has said that they aren't planning to introduce themes support at the moment. I came up #947 recently and its ability to invert colors and perform per-channel gamma correction might help here. A pre-compiled firmware is available at https://tiplanet.org/forum/viewtopic.php?f=97&t=22485 if you want to try it out.

Otherwise, changing the data color palette would be the way to go (escher/include/escher/palette.h).

DSchndr:
Yeah, i also know about the theme support, but i think that it would be awesome to add some accessibility options to epsilon to make the calculator more interesting.
Per-channel gamma correction would be really nice, because then we don't need to correct images if there are any that need correction.

As for the graphing issue, i've replaced the palette and it seems to work, but i still need to ask how much better it looks.

```
//Paul Tol colorblind pallette (https://davidmathlogic.com/colorblind)
  constexpr static KDColor CBPurple = KDColor::RGB24(0x332288);
  constexpr static KDColor CBGreen = KDColor::RGB24(0x117733);
  constexpr static KDColor CBOGreen = KDColor::RGB24(0x44AA99);
  constexpr static KDColor CBBlue = KDColor::RGB24(0x88CCEE);
  constexpr static KDColor CBYellow = KDColor::RGB24(0xDDCC77);
[...]

PR #1385, opened 2020-03-06
Manually change the color of functions/sequences

### LINKED ISSUE #1247: Manually change the color of functions/sequences
Currently the color of the functions and sequences we plot are random.
I think it would be a good to add the option to  **change the color of the functions** we plot in the function app :) (and also in the sequence app ofc)

- Like for instance if we want to differentiate derivatives from their primitives, it would be easier to differentiate them.

- Or also if you want to create 1 function out of 2/3/... functions (a function defined by parts), then put them both at the same color ;)

- etc.

In direct acces in this menu : 
screenshot (5)
screenshot (6)

Thank you for reading :)

EbbDrop:
Adds a new menu to the function/sequence option-menu to change the color.
Screenshot_from_2020-03-06_01-55-25
Old version see update below!
Fixes #1247 and helps with #931


MixedMatched:
I would prefer adding a color_view to escher and using gauges as well as presets. That way colors don't have a limited amount. I'll see what I can do later today.

EbbDrop:
I will add a custom color option. But I don't think it is necessary to make a color_view just for functions.

MixedMatched:
How are you going to show what color something is in the preference menu if you can't show color? Are you just going to say "Custom"? And, if we use a color_view, we can create a row (not a column) of small cells that just have colors instead of a list of colored words.

EbbDrop:
Sorry i think i misunderstood you. A color_view seems like a good idee. How would you implement the custom color option?

EbbDrop:
Added a nicer color selector by making a color_view:

It also works with longer names:

MixedMatched:
Wow! That's a very clean implementation. I'm going to see if I can add a clean way for custom colors. If I make one, I'll PR your PR in a few hours.

cent20net:
I love this feature ! Smart and easy to use !

EbbDrop:
@cent20net Thanks :)

artaxxx:
Thanks @EbbDrop , that's interesting :)

I think the list display is better because it's more consistent with the interface of the product. Here are the mockups we worked on to fit the UI:

metawops:
Still no coloring option in Grapher in the current version (18.2.0), right? :-(((

Yaya-Cout:
It's in the 19.0.0 beta version

metawops:
> It's in the 19.0.0 beta version.

Awesome! Unfortunately, can’t install it on my device. Tried several times. Stays at 18.2.0. 🤷‍♂️

artaxxx:
> > It's in the 19.0.0 beta version.
> 
> Awesome! Unfortunately, can’t install it on my device. Tried several times. Stays at 18.2.0. 🤷‍♂️

The beta version is not available for N0100 models for now

metawops:
> > > It's in the 19.0.0 beta version.
> > 
> > 
> > Awesome! Unfortunately, can’t install it on my device. Tried several times. Stays at 18.2.0. 🤷‍♂️
> 
> The beta version is not available for N0100 models for now

Ahhh, thanks for the hint! It‘s very „unfortunate“ that the process seems to finish successfully, though …
4C7D2456-5F7E-4783-9829-AA87DAEC98C3

So, what other models are there? I, indeed, have a N0100 version … 😞

PR #1400, opened 2020-03-13
Fix simplification and approximation2

LeaNumworks:
This fixes the computation of 1.0092^50*ln(3/2), that returned 0 due to an overflow when computing a LCM.


LeaNumworks:
There was a separate problem for `1.0092^50*ln(1.0092)` (multiplication overflow, now fixed)

Issue #1311, opened 2020-02-17
Navigation buttons are always visible


Ecco:
#### Describe the bug
Building the latest version using the latest Android SDK yields an app where the navigation buttons are always visible.

#### Screenshots
![screen](https://user-images.githubusercontent.com/55989/74660148-944e6800-5163-11ea-9bc5-4f99ac67c896.png)

#### Expected behavior
The navigation bar at the bottom should auto-hide. Note that it work well on the very same phone using the version currently on Google Play.

#### Environment
 - Epsilon master
 - Running on Android 10
 - Built using latest Android SDK

#### Android SDK details

  Path                 | Version      | Description                    | Location             
  -------              | -------      | -------                        | -------              
  build-tools;28.0.2   | 28.0.2       | Android SDK Build-Tools 28.0.2 | build-tools/28.0.2/  
  ndk-bundle           | 21.0.6113669 | NDK                            | ndk-bundle/          
  patcher;v4           | 1            | SDK Patch Applier v4           | patcher/v4/          
  platform-tools       | 29.0.5       | Android SDK Platform-Tools     | platform-tools/      
  platforms;android-29 | 4            | Android SDK Platform 29        | platforms/android-29/
  tools                | 26.1.1       | Android SDK Tools 26.1.1       | tools/

artaxxx:
Fixed by #1410

PR #1410, opened 2020-03-23
[ion/android] Hide the status bar and navigation buttons

LeaNumworks:
This is a hack, we should find why this was broken after v12.


Ecco:
#1311

PR #1412, opened 2020-03-24
Fix poincare crash

EmilieNumworks:
- Avoid infinite loop : a^x*b^x --> (a¨b)^x --> a^x*b^x --> (a^b)^x when a and b are very big integers
- Avoid approximating 1.0E+92^(-0.5) --> 0.0

### DIFF
--- a/poincare/src/expression.cpp
+++ b/poincare/src/expression.cpp
@@ -321,18 +321,22 @@ void Expression::defaultDeepReduceChildren(ExpressionNode::ReductionContext redu
 
 Expression Expression::defaultShallowReduce() {
   Expression result;
-  const int childrenCount = numberOfChildren();
-  for (int i = 0; i < childrenCount; i++) {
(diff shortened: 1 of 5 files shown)


poincare/src/expression.cpp
```diff
@@ -321,6 +321,9 @@ void Expression::defaultDeepReduceChildren(ExpressionNode::ReductionContext redu
 
 Expression Expression::defaultShallowReduce() {
   Expression result;
+  if (sSimplificationHasBeenInterrupted) {
+    result = Undefined::Builder();
+  }
   const int childrenCount = numberOfChildren();
```
LeaNumworks: Put the following lines in a else?

EmilieNumworks: I did not want to make the code reading more difficult but you're right, it does slightly optimise the execution. Trade-off?

PR #1416, opened 2020-03-25
[ion/emscripten] Use pointer events instead of mouse events

LeaNumworks:
Otherwise, clicks are not handled on mobile devices.


Ecco:
See #1418

PR #1419, opened 2020-03-26
[Feature] Shift + Home is now a shortcut to go to calculation

0b101:


Ecco:
Thanks for this PR @0b101 ! That's an interesting suggestion.

This might be useful for a power user, but I'm concerned about a false positive that would come as a surprise for a novice user. For instance, somebody typing shift-home without noticing shift was active, would be sent straight to the Calculation app instead of the Home app as expected.

0b101:
@Ecco Do you think this would work better under some sort of setting that also enables a few other shortcuts like shift + ( for (), shift + +/- for brightness control, and stuff like that?

Ecco:
As a general rule we try to avoid adding settings as much as we can.

That being said, I'm all for extra features as long as they don't disturb newcomers. Having a shortcut to go directly to the Calculation app makes sense, and so would having a shortcut for brightness control. I think the best solution would simply to find a shortcut that is very unlikely to be accidentally triggered by   an unsuspecting user.

And I'm not saying that "Shift-Home" is a bad idea: I'm just wondering *if* it could accidentally disturb a beginner.

0b101:
I’ll ask around to see what combo people would assume to send them to the calculation app. I’ll also try to find some newer users and I’ll ask them about an unexpected trip to the calculation app. I’m guessing that they’ll probably just assume that they pressed the “OK” button after home, but it’s safe to be sure.

artaxxx:
Good idea to give a shortcut to the Calculation app! All things considered, setting this shortcut to the Home key is probably not too much of a problem for newcomers. In fact, pressing Home already indicates a willingness to exit the current application. If arriving in the Calculation app is too surprising, you can just press Home again.

It would be interesting to merge this PR.

However, there's a bug in this PR @0b101: if I press shift+Home in the Calculation app, the banner at the top of the screen should not display "shift". As well as pressing shift+OK in the calculation app: it does nothing but "shift" disappears from the top banner.

0b101:
> However, there's a bug in this PR @0b101: if I press shift+Home in the Calculation app, the banner at the top of the screen should not display "shift". As well as pressing shift+OK in the calculation app: it does nothing but "shift" disappears from the top banner.

@artaxxx It appears that the "shift" text is actually phantom text. I fixed it in the latest commit by sending the calculator to home before calculations.

artaxxx:
Thanks @0b101 ! I flashed a N0110 and there's a problem when you press shift+Home: before the Calculation screen, the Home screen appears during a microsecond. Do you see that?

0b101:
That must have to do with how I cleared the “shift” text. I’ll have to find another solution. I tried using the shift alpha clear function, but it did nothing, so I sent the calculator to the home app before calculation.

apps/apps_container.cpp
```diff
@@ -185,6 +185,11 @@ bool AppsContainer::processEvent(Ion::Events::Event event) {
     switchTo(appSnapshotAtIndex(0));
     return true;
   }
+  if (event == Ion::Events::ShiftHome) {
+    switchTo(appSnapshotAtIndex(0)); // fixes phantom "shift" text
+    switchTo(appSnapshotAtIndex(1));
```
RedGl0w, line 190:
```suggestion
    switchTo(appSnapshotAtIndex(1));
    m_window.redraw(true);
```

0b101:
@artaxxx Thanks for pointing that out. It should be fixed now with RedGl0w’s suggestion.

RedGl0w:
My solution is not perfect, because I think it redraws the hole screen, so maybe the history could also blink. I'm trying to only redraw the top bar

Ecco:
It's weird that we have to `m_window.redraw(true);`. The root cause for this problem is probably somewhere else since this patch is kinda simple.

RedGl0w:
> Good idea to give a shortcut to the Calculation app! All things considered, setting this shortcut to the Home key is probably not too much of a problem for newcomers. In fact, pressing Home already indicates a willingness to exit the current application. If arriving in the Calculation app is too surprising, you can just press Home again.

Do you think having 789456123 keys to access apps would be a big issue for newcomers ?
It would make all changes of app way faster for experimented user
And would be pretty evident once you saw it...

artaxxx:
Well, having 789456123 keys to access apps is a big issue for everyone... I'm not sure we want to publish a dictionary to list all the shortcuts 😅

Now, seriously, it all depends on the shortcut:
* the shortcut has to be meaningful (example: don't use the `sqrt` key to turn off the calculator)
* the shortcut must not interfere with the normal use of the calculator (example: don't use `shift`+`→` to adjust the brightness of the screen)
* if a newcomer use the shortcut by mistake, it has to be simple to undo the last action (the example of this PR is good, it's precisely a shortcut that allows the user to easily undo the last action!)

RedGl0w:
> Well, having 789456123 keys to access apps is a big issue for everyone... I'm not sure we want to publish a dictionary to list all the shortcuts 😅

I meant 7 was for top left app, 8 for top middle,... 
It can be Un-done easily, and it's very usefull

artaxxx:
Well, OK then. 
It's not that obvious. For instance, Casio uses a different system (1 is for the first app, 2 for the second, ...) and what happens the day we add a new app?

RedGl0w:
> Well, OK then.
> It's not that obvious. For instance, Casio uses a different system (1 is for the first app, 2 for the second, ...) and what happens the day we add a new app?

yep that's the main issue of these shortcuts
But currently, epsilon is made so we have to go to a specific app to do something specific (to plot a graph, you'll have to go to graph, to calculate the max of a function, you'll have to go to it too), and "menuing" is very long on epsilon currently.

I have another "shortcut" (but way less complicated) idea :
In the calculation history, when we press an operator key (for instance x²), a new calculation could be added in the history, which contains the sectioned one with the operator

Ecco:
I like the idea of your last proposal though! But that's a different topic. Would you mind opening another issue so we can discuss this one separately?

RedGl0w:
> I like the idea of your last proposal though! But that's a different topic. Would you mind opening another issue so we can discuss this one separately?

Okay 👌

PR #1422, opened 2020-03-31
[apps/solver] Fix EquationStore::resolveLinearSystem

LeaNumworks:
Scenario: solve a system with the maximum number of variables -> our
routine to compute the number of variables did not use
k_maxNumberOfVariables, so we ended up adding the same children to the
matrix Ab, which created ghosts, which made the deep reduction crash.

### DIFF
--- a/apps/solver/equation_store.cpp
+++ b/apps/solver/equation_store.cpp
@@ -211,7 +211,16 @@ EquationStore::Error EquationStore::privateExactSolve(Poincare::Context * contex
     // Step 3. Polynomial & Monovariable?
     assert(numberOfVariables == 1 && numberOfDefinedModels() == 1);
     Expression polynomialCoefficients[Expression::k_maxNumberOfPolynomialCoefficients];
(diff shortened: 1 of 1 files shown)


apps/solver/equation_store.cpp
```diff
@@ -262,7 +271,9 @@ EquationStore::Error EquationStore::resolveLinearSystem(Expression exactSolution
   Preferences::AngleUnit angleUnit = Preferences::sharedPreferences()->angleUnit();
   // n unknown variables
   int n = 0;
-  while (m_variables[n][0] != 0) { n++; }
+  while (m_variables[n][0] != 0 && n < Expression::k_maxNumberOfVariables - 1) {
+    n++;
+  }
```
EmilieNumworks: Shouldn't it be:
'while (n < Expression::k_maxNumberOfVariables && m_variables[n][0] != 0) {
     n++;
   }'

LeaNumworks: Indeed, thanks!

PR #1436, opened 2020-04-03
Refactor the build system

Ecco:
- Use variants (e.g. `src += foo/bar.cpp:+variant`)
- macOS and iOS can now build any binary (e.g. `make test.app`)
- Streamline the root Makefile
- Share multi-arch Make code
- Get rid of DEFAULT target in favor of first HANDY_TARGET
- Add extra (and cheap) targets to the CI


Ecco:
Make sure to compare the results of all build targets before merging (including bench targets).

build/targets.simulator.ios.mak
```diff
@@ -0,0 +1,10 @@
+ifndef ARCH
+HANDY_TARGETS_EXTENSIONS += apk
```
EmilieNumworks, line 2: ipa?

Ecco: Well, not even that. Those are in `build/platform.*.mak` now 😄

Ecco:
- Comparing all_official seems ok (minor differences, at least due to the commit being embedded). By the way, the all_official target is broken. PR incoming.
- Comparing binpacks seems ok too (again, super small differences, a few bytes).

RedGl0w:
@Ecco it seems that moving the `SFLAGS` to default.mak is definitely not a good idea... Now, even when you just compile the linux simulator, the `EPSILON_GETOPT` is true, and so it causes https://github.com/numworks/epsilon/issues/1493

PR #1439, opened 2020-04-03
Added some Imperial units

0b101:
UNITS

I added some Imperial translations for distance and mass. There are currently no translations.


Ecco:
Hi @0b101 ! That's a great idea, thanks! However as you can see it would require a lot more work. I'm going to close this PR since it has no chance of working as-is. Feel free to reopen or create a new one if needed.

PR #1446, opened 2020-04-05
Homogénéisation avec variables.fr.i18n

XanderLeaDaren:


EmilieNumworks:
Thanks @XanderLeaDaren for this suggestion.
@artaxxx: the PR concerns only French, should it be extended to other languages as well?

artaxxx:
Thanks @XanderLeaDaren ! Emilie is right, could you please extend the changes to other languages?

Ecco:
I'm sorry but I don't really get the point of this PR. Using a dash is a common way to designate a character range, which is used in [POSIX RegEx](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended) for example.

Is there something I'm missing?

PR #1448, opened 2020-04-05
Relecture

XanderLeaDaren:


EmilieNumworks:
Thanks @XanderLeaDaren, the trailing space after "Intervalle" is indeed useless! However, I let @artaxxx decides regarding the spelling of "Boite"/"Boîte".

artaxxx:
Thanks @XanderLeaDaren ! I think we'll keep "boîte" for now and merge the other half. Could you please update your PR and add the same change for the English version? There's also a useless space after "interval".

PR #1450, opened 2020-04-05
Build size analysis

Ecco:
This PR adds a tool to compare the size of two binaries, and hooks it into the CI system to report it on any incoming PR.

### DIFF
--- a/build/metrics/binary_size.py
+++ b/build/metrics/binary_size.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import re
+import subprocess
+import urllib.parse
+
+# ELF analysis
+
+def loadable_sections(elf_file, address_prefix = ""):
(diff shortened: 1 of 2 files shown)


Ecco:
Note: Current failure is most likely due to running from a PR (and therefore not having the proper auth token). Should be OK once merged.

build/metrics/binary_size.py
```diff
@@ -0,0 +1,140 @@
...
+import argparse
+import os
+import re
+import subprocess
+import urllib.parse
+
+# ELF analysis
+
+def loadable_sections(elf_file, address_prefix = ""):
+  objdump_section_headers_pattern = re.compile("^\s+\d+\s+(\.[\w\.]+)\s+([0-9a-f]+)\s+([0-9a-f]+)\s+("+address_prefix+"[0-9a-f]+)\s+([0-9a-f]+).*LOAD", flags=re.MULTILINE)
```
LeaNumworks, line 12: Maybe add a comment to explicit this to the reader?

Ecco: Well… this is an unfair request :-D This exact same code already made it to the tree without any comment whatsoever. See `build/device/elf2dfu.py` 😄

RedGl0w:
It doesn't work well :/
See https://github.com/numworks/epsilon/pull/1459/checks

PR #1454, opened 2020-04-06
Fix default language

LeaNumworks:
This PR fixes the problem that was spotted by #569


0b101:
@LeaNumworks I don't know if this is intentional, but the simulator now defaults to French instead of English

RedGl0w:
@LeaNumworks there's an issue now with theses lines :
https://github.com/numworks/epsilon/blob/4a9bfca1f6a02edfcea5fb5068891f48bfe309e7/apps/main.cpp#L30-L43
This part should be totally refractored, because it doesn't mind about the EPSILON_I18N...

Issue #1371, opened 2020-02-29
type(enumerate()) Python code makes the calculator crash


sagessylu:
Executing the code `type(enumerate())` in the Python app makes the calculator crash.
This is an issue for both Python scripts and Python shell.
However, it works just fine if you put something between the brackets of `enumerate()`.

![screenshot](https://user-images.githubusercontent.com/49091098/75606029-14bc8380-5ae9-11ea-9c0c-c00d17a65cf0.png)
![screenshot(1)](https://user-images.githubusercontent.com/49091098/75606053-3d447d80-5ae9-11ea-9272-999b1676e0c4.png)

#### To Reproduce
1. Go to the Python app
2. Go to the Python shell
3. Type `type(enumerate())`
4. ⚠️ The calculator crashes

#### Expected behavior
It should print the same error as when you just run `enumerate()`

#### Environment
 - Epsilon version : 12.4.0
 - Works only on the real calculator (not the online simulator)

RedGl0w:
Seems to be patched with mpy 12 (tested with mpy 1.12.0)

EmilieNumworks:
Good catch @sagessylu. The bug is caused by a invalid memory reading; in `enumerate_make_new, we access `args[0]` even when `n_args = 0`... This was fixed in MicroPython source by this [https://github.com/micropython/micropython/commit/4ebbacd65eec4c0fde8ceb32274c06f5e3de3af6](commit) included in the version 1.12 (thanks @RedGl0w). We should definitely update the Micropython version embedded in Epsilon!

RedGl0w:
@EmilieNumworks this can be pached so with #1455 !

EmilieNumworks:
Done in #1457!

Ecco:
Well, #1457 hasn't been merged just yet…

EmilieNumworks:
Right, wrong button 😅

Issue #1356, opened 2020-02-24
Upgrade MicroPython from 1.11 to 1.12


0b101:
MicroPython is out of date. It should be upgraded to allow new features and bug fixes to come to the calculator. In addition to that, upgrading should save around 584 bytes. For more information, see: https://github.com/micropython/micropython/releases

EmilieNumworks:
Done in #1457!

PR #1455, opened 2020-04-06
[python] micropython: update to v1.12-326-g8fff0b0ac-dirty

KaiJan57:
This pull request updates micropython. Mostly untested, but basic functionality is guaranteed.
Obviously fixes this [issue](https://github.com/numworks/epsilon/issues/1356) and apparently fixes this [issue](https://github.com/numworks/epsilon/issues/1371)


EmilieNumworks:
Very good idea @KaiJan57. However, we would rather embed the latest release v1.12 of MicroPython instead of the current head of master (which might be unstable). The PR #1457 does that and should be merged soon.

Issue #1231, opened 2019-12-04
Error in build documentation for Linux: ./epsilon.elf doesn't exist after build


mobluse:
#### Describe the bug
There is an error in the documentation at https://github.com/numworks/epsilon/blob/master/docs/build/index.md
```
$ ./epsilon.elf
bash: ./epsilon.elf: No such file or directory
```
However this works and starts the NUMWORKS simulator:
`$ ./output/release/simulator/linux/epsilon.bin
`

#### Screenshots
N/A

#### To Reproduce
See description above.

#### Expected behavior
The NUMWORKS simulator should start.

#### Environment
 - Epsilon version (Settings > About > Software version): 12.0.0 (I cloned GitHub today so it's strange that this version number is lower than the version for Android that was released earlier: 12.3.0. Is the GitHub source not the latest source?)
 - The platform(s) on which the problem happens: online simulator, actual device, etc...
NumWorks simulator in Raspbian Buster Linux running on Raspberry Pi 4 B.

Ecco:
Indeed. I think we could use a specific target. Maybe `make epsilon_run`?

HollMax:
Just encountered this issue myself. I'd start by updating the documentation to reflect what's happening rather than adding a new target.

RedGl0w:
> Just encountered this issue myself. I'd start by updating the documentation to reflect what's happening rather than adding a new target.

Not a good idea : on macos, it's not in the same location...
It would be more complicated...

RedGl0w:
Closed with https://github.com/numworks/epsilon/commit/80f3af389fc1247c65b124e3f2d7ff83eb6a1e9e#diff-44575cf5b28512d75644bf54a517dcef304ff809fd511747621b4d64f19aac66

RedGl0w:
> Indeed, thanks for the bug report! I think we could use a specific target. Maybe `make epsilon_run`?

This is what was added with epsilon 15 :p

LeaNumworks:
This issue is indeed outdated. Thanks!

PR #1468, opened 2020-04-07
Added make epsilon_run and make cleanall

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1231#issuecomment-562140084

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,11 @@ clean:
 	@echo "CLEAN"
 	$(Q) rm -rf $(BUILD_DIR)
 
+.PHONY: cleanall
+cleanall:
+	@echo "CLEANALL"
+	$(Q) rm -rf output
+
 .PHONY: cowsay_%
 cowsay_%:
 	@echo " -------"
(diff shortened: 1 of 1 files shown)


EmilieNumworks:
Hi @RedGl0w, the %_run target should be platform dependent.
See in build/target.device.mak, it's already implemented:

```
.PHONY: %_run
%_run: $(BUILD_DIR)/%.$(EXE)
	$(GDB) -x build/$(PLATFORM)/gdb_script.gdb $<
```

or in build/target.simulator.android.mak or build/target.simulator.ios.mak:

```
.PHONY: %_run
%_run: $(BUILD_DIR)/%.apk
	$(call rule_label,ADB)
	$(Q) adb install $<
```

Your implementation works for windows I believe. On macos, your would want to do something like
`open output/debug/simulator/macos/epsilon.app`.

RedGl0w:
> Hi @RedGl0w, the %_run target should be platform dependent.

That is exactly what I'm trying to not do :trollface: 
Like Ecco said here : https://github.com/numworks/epsilon/issues/1231#issuecomment-562140084, a crossplatform solution to start the simulator could be very usefull. On linux it works well, I'm trying on macos...

EmilieNumworks:
But it won't work on Android, iOS nor on the device?

Ecco:
> That is exactly what I'm trying to not do :trollface:

But that's exactly what's going to **not** work 😄

Ecco:
Also, this PR does two things. Not a big problem, but it'd be much better to open two PRs so we could discuss each feature individually.

RedGl0w:
> Also, this PR does two things. Not a big problem, but it'd be much better to open two PRs so we could discuss each feature individually.

okay, I'll try to do this today

PR #1478, opened 2020-04-08
[build/simulator] NDEBUG flag in release mode

LeaNumworks:
This way, asserts are not performed on simulator platforms


Ecco:
Great 👍

Issue #1186, opened 2019-11-06
percentage placement on N0110


AlexPMLikesNumWorks:
Problem:
The N0110 has now a percentage sign in alpha mode. instead of the backspace character. This makes editing very awkward and counterintuitive.

Suggestion:
Why not place the % on top of the .10^x key, still unused in alpha mode and much much less useful than the backspace in alpha mode?
At least make it configurable, so as to allow conservation of backspace capability while typing in alpha mode if the user prefers. This is much more important than the accessibility of the % symbol.

parisseb:
I have moved it to shift-division key in the Delta firmware and restored backspace as backspace in alpha mode too. The position is easy to remind because % is integer division remainder in Python.

artaxxx:
It is convenient to keep the % sign on the keyboard. So we finally decided to disable it only in "alpha lock" mode to avoid this editing problem.

RedGl0w:
> It is convenient to keep the % sign on the keyboard. So we finally decided to disable it only in "alpha lock" mode to avoid this editing problem.

At least, the epsilon users will be used to omega :trollface: 
https://github.com/Omega-Numworks/Omega/pull/101/files

RedGl0w:
> It is convenient to keep the % sign on the keyboard. So we finally decided to disable it only in "alpha lock" mode to avoid this editing problem.

Is it already done ?

RedGl0w:
This issue can now be closed with #1479 merged

PR #1479, opened 2020-04-08
Changed percent use in alpha lock

RedGl0w:
Like it was said here : https://github.com/numworks/epsilon/issues/1186#issuecomment-610412787
Based on : https://github.com/Omega-Numworks/Omega/pull/101/files

Now the backspace key in alpha lock just remove the last glyph, but in alpha add a `%`

### DIFF
--- a/escher/src/layout_field.cpp
+++ b/escher/src/layout_field.cpp
@@ -473,6 +473,11 @@ bool LayoutField::privateHandleEvent(Ion::Events::Event event) {
     if (!isEditing()) {
       setEditing(true);
     }
+
+    if (event == Ion::Events::Percent && Ion::Events::isAlphaLockActive()) {
+      event = Ion::Events::Backspace;
+    }
+
     if (event.hasText()) {
       handleEventWithText(event.text());
     } else if (event == Ion::Events::Paste) {
--- a/escher/src/layout_field.cpp
+++ b/escher/src/layout_field.cpp
@@ -473,11 +473,6 @@ bool LayoutField::privateHandleEvent(Ion::Events::Event event) {
     if (!isEditing()) {
       setEditing(true);
     }
-
-    if (event == Ion::Events::Percent && Ion::Events::isAlphaLockActive()) {
-      event = Ion::Events::Backspace;
-    }
-
     if (event.hasText()) {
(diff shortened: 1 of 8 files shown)


Ecco:
That being said, definitely a behavior we'll want to merge.

RedGl0w:
> That's plain code duplication. We'd need to avoid this.

Ok I'm working on a new version in events.cpp

RedGl0w:
@Ecco what do you think is best :
- change s_dataForEvent to PageSize (in ion/includes/ion/keyboard/.../layout.h), and so have a very clean solution, but which will be a bit heavy useless
- add in `Event::Event(Keyboard::Key key, bool shift, bool alpha)` a condition about if alphalock is actived to map to another event
?

Ecco:
I also think the second option would be best. That being said, are we sure that we will *always* want to return a Backspace event for Alphalock-Percent? Is there any case where we'd rather get the percent? Because otherwise we'd still need to dispatch a Percent event and handle this further down the road just like you did initially (but factor the event-replacement logic).

RedGl0w:
> That being said, are we sure that we will _always_ want to return a Backspace event for Alphalock-Percent? I would say yes, but is there any case where we'd rather get the percent? Because otherwise we'd still need to dispatch a Percent event and handle this further down the road just like you did initially (but factor the event-replacement logic).

I think that we shouldn't think like that. An universal solution is better than a solution which gives too much possibilities, and make the user lost...

RedGl0w:
Ok I updated with a "universal" solution.

RedGl0w:
Oh, maybe I broke the tests,...
I can't see the github actions on mobile, so I'll try to understand tomorrow what's going on

RedGl0w:
This solution is quite easy, but works. The tests are updated. Now, feel free to choose if the changes you've requested, @Ecco are good now...

Ecco:
Looks good to me 😄Needs some manual testing before being merged.

RedGl0w:
Should I squash the commits ? (it would be more beautifull in history, but this conversation would be incomprehensible...)

Ecco:
I gave this a spin and it looks just great, so I just (squash-)merged it 😄

PR #1484, opened 2020-04-09
[Calculation] Allow inserting a past result at the *denominator* of a fraction

Ecco:
1. Go to the Calculation app
2. Type `123`
3. Now type `456/` and try retrieving the `123` above to make `456/123`.

### DIFF
--- a/apps/calculation/expression_field.h
+++ b/apps/calculation/expression_field.h
@@ -7,7 +7,10 @@ namespace Calculation {
 
 class ExpressionField : public ::ExpressionField {
 public:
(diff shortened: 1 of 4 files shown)


Ecco:
**Note**: Don' merge as-is, this creates a glitch when creating a function for the second time in the Functions app (extraneous margin on the left).

LeaNumworks:
> **Note**: Don' merge as-is, this creates a glitch when creating a function for the second time in the Functions app (extraneous margin on the left).

This glitch was not created by this PR, and is fixed by #1546

PR #1487, opened 2020-04-09
[Python] Added a script size cell

RedGl0w:
Rebase of : https://github.com/numworks/epsilon/pull/1349

In the script option, there's just a new cell, and when you press ok, it just switch the type of display (percent <-> x o / storagelength o).

Screens (in every language) :
image
image
image
image
image
image


artaxxx:
I have given the question a bit of thought and finally I think that this information should not appear in the firmware.

This is a nice feature but it provides technical information that is not really necessary. Finally, what is the point of knowing the size of a specific script? What is interesting is to know the remaining global storage space. The problem is that the remaining space is also related to the functions and variables stored in memory. It's not just about python.

Actually, we're talking about the possibility of having a file explorer. This should be the best thing to indicate this information. But it's not something we want to do in the first place.

RedGl0w:
> Actually, we're talking about the possibility of having a file explorer. This should be the best thing to indicate this information. But it's not something we want to do in the first place.

What do you mean by "file explorer" ?
An app to show all records, and extension ?
A new version of Ion::Storage with file tree ?

artaxxx:
Yes but I'm not even convinced it's relevant.

RedGl0w:
> Yes but I'm not even convinced it's relevant.

If there's a file tree, this could be very usefull for scripts made of multiple script, for having a clean python script list.

artaxxx:
Sure, it's cool, but it adds a lot of complexity. That's why we're not thinking of adding this kind of feature.

PR #1490, opened 2020-04-09
CI windows

EmilieNumworks:
- make sure that a test failure is detected by the CI on windows platform
- fix some tests on windows


Ecco:
I'm curious why `msys2do` is required to properly run the test executable. The whole point of the mingw toolchain is to generate *native* windows binaries.

PR #1494, opened 2020-04-09
[poincare] Get rid of ArrayBuilder

Ecco:


poincare/include/poincare/expression.h
```diff
@@ -286,6 +285,9 @@ class Expression : public TreeHandle {
 
   static void Tidy() { sSymbolReplacementsCountLock = false; }
 
+  /* Tuple */
+  typedef std::initializer_list<Expression> Tuple;
```
EmilieNumworks, line 289: I find it clearer with `std::initializer_list<Expression>` or `std::initializer_list<TreeHandle>` instead of Tuple

Ecco: I don't 😄

Issue #1493, opened 2020-04-09
[Bug] The simulator defaults to French instead of English


0b101:
#### Describe the bug
In #1454 it mentions making the first listed language the default, but it defaults to the second listed language instead,

Here is the list from build/config.mk: `EPSILON_I18N ?= en fr es de pt` 

#### Expected behavior
The first listed language in build/config.mk should be the default.

RedGl0w:
There's definitely something not working well : the default language when I compile is es...

RedGl0w:
It seems that the ion_main with args (with language) is called, even when there's no args : by commenting these lines https://github.com/numworks/epsilon/blob/4a9bfca1f6a02edfcea5fb5068891f48bfe309e7/apps/main.cpp#L38, the language is the right one. I think this comes from another PR.

LeaNumworks:
Hi @0b101, thanks for reporting this bug!
Should be fixed by #1496.

PR #1495, opened 2020-04-10
EPSILON_GETOPT patch with default language

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1493.
Like I said there https://github.com/numworks/epsilon/pull/1454#issuecomment-611909791, the GETOPT for language should be refractored...


RedGl0w:
Closed with https://github.com/numworks/epsilon/pull/1496

LeaNumworks:
Hi @RedGl0w, thanks for your PR!
As you saw, I did PR #1496, which fixes the bug in a cleaner way by getting the languages from the i18n, as you suggested.

RedGl0w:
> Hi @RedGl0w, thanks for your PR!
> As you saw, I did PR #1496, which fixes the bug in a cleaner way by getting the languages from the i18n, as you suggested.

Yes, that's **way** cleaner

PR #1496, opened 2020-04-10
[apps] GET_OPT uses generated i18n language codes to find language

### LINKED ISSUE #1493: [Bug] The simulator defaults to French instead of English

LeaNumworks:
Fixes #1493
- We removed Default from the Languages list in #1454, which invalidated our algorithm to parse the argument "--language" in apps/main.cpp, which resulted in the bug described in #1493
- This PR cleans said algorithm, so we do not need the dangerous `const char * languageIdentifiers[]` anymore

I added i18n files for the ISO639-1 codes for languages, but those i18n are only built when `EPSILON_GET_OPT=1`, so not for the device

Maybe we should remove the "none" case?

### DIFF
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -57,6 +57,9 @@ $(call object_for,apps/apps_container_storage.cpp apps/apps_container.cpp apps/m
 SFLAGS += -I$(BUILD_DIR)
 
 i18n_files += $(addprefix apps/language_,$(addsuffix .universal.i18n, $(EPSILON_I18N)))
+ifeq ($(EPSILON_GETOPT),1)
+i18n_files += $(addprefix apps/language_,$(addsuffix _iso6391.universal.i18n, $(EPSILON_I18N)))
+endif
 i18n_files += $(addprefix apps/,\
   shared.de.i18n\
   shared.en.i18n\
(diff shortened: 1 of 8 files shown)


RedGl0w:
> 	.text.external 	.rodata.external 	Total
> **Base** 	633 840 bytes 	231 583 bytes 	877 99 bytes
> **Head** 	633 864 bytes 	231 591 bytes 	877 131 bytes
> 	_+ 24 bytes_ 	_+ 8 bytes_ 	_+ 32 bytes_
> 	_+ 0 %_ 	_+ 0 %_ 	_+ 0 %_

Is the GETOPT true on device ? Because it shouldn't be heavier :thinking:

LeaNumworks:
> Is the GETOPT true on device ? Because it shouldn't be heavier 🤔

The PR is ok, I just had not rebased it just before pushing it, so the bot comparison was not accurate :)

Issue #1489, opened 2020-04-09
Add the turtle defaults colors to the kandinsky module in Python app


sagessylu:
#### Enhancement idea

It would be nice if you could use for example "orange" as a correct setting for the kandinsky module, just like for the turtle module.

#### Screenshot

Just like this :
![screenshot](https://user-images.githubusercontent.com/49091098/78898989-04e37800-7a75-11ea-80aa-6594fe1283c1.png)

sagessylu:
@RedGl0w : Does this mean that this turtle feature I'd like to have in kandinsky is not a standard feature ? Oh... :cry:

RedGl0w:
Kandinsky is not a Standart feature, so your idea is possible. I'm currently working on it

RedGl0w:
It will be fixed with https://github.com/numworks/epsilon/issues/1533

PR #1503, opened 2020-04-11
Python kandinsky supports string as colors

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1489


Ecco:
Hi @RedGl0w ! Thanks a lot for this PR! This is definitely a great addition, and we'd like to merge something very similar albeit subtly different.

I wrote an extensive definition on what we'd want to merge in #1533 .

As a result I'm going to close this PR, but feel free to:
 * Discuss the specification in #1533 
 * Create a new PR that matches #1533 

Thanks!!

RedGl0w:
>     * Create a new PR that matches #1533

That's exactly what I'm going to work on in the next days !

PR #1512, opened 2020-04-12
Real random with emscripten

RedGl0w:
So we can have **real** random with emscripten

### DIFF
--- a/ion/src/shared/random.cpp
+++ b/ion/src/shared/random.cpp
@@ -1,11 +1,18 @@
 #include <ion.h>
 #include <stdlib.h>
+#if __EMSCRIPTEN__
+#include <emscripten.h>
+#endif
 
 uint32_t Ion::random() {
   /* rand() returns a pseudo-random integral number in the range between 0 and
    * RAND_MAX. We assert that RAND_MAX is at least 8 bit, this way we can merge
    * four rand() calls to create a single 32 bit value. Merging is done using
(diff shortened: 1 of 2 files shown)


EmilieNumworks:
`emscripten_random` is implemented using `Math.random()` of javascript standards whereas we used `rand()` of the C++ standard library. I am not familiar with both the implementations, why do you think one is better than the other?
The issue you might have collaterally resolved  is that the seed used for emscripten seems to be always the same... It had been resolved [there](https://github.com/numworks/epsilon/commit/8e725ce328a41482935361b621a74afc784fa4cb) but is broken again. I'll look into it now.

RedGl0w:
> `emscripten_random` is implemented using `Math.random()` of javascript standards whereas we used `rand()` of the C++ standard library. I am not familiar with both the implementations, why do you think one is better than the other?
> The issue you might have collaterally resolved is that the seed used for emscripten seems to be always the same... It had been resolved [there](https://github.com/numworks/epsilon/commit/8e725ce328a41482935361b621a74afc784fa4cb) but is broken again. I'll look into it now.

That was exactly for this reason, but I think using `Math.random()` is better than time for setting the seed. I'll commit that in a few minutes

RedGl0w:
Closed by https://github.com/numworks/epsilon/pull/1517

EmilieNumworks:
The feature is broken on other platforms (ios, android) as well. So we want to fix it for all at once - see #1517.

PR #1514, opened 2020-04-13
Use std::min and std::max

Ecco:
This removes an *awful* lot of duplicated code…

Note: Sometimes the templated function (either `std::min` or `std::max`) needs to be explicited (e.g. `std::min<float>(foo, bar)`). This is undesirable, and may be a result of either:

* Unexpected type promotion : For instance, given `KDCoordinate length = 2;` the type of `2*length` will be `int`, not `KDCoordinate`. Maybe it could be a good idea to promote `KDCoordinate` to `int`? It would fix this, and might be a bit faster (albeit a tad bit bigger).
* Floating point mismatch: some function mix `double` and `float` for no apparent reason (e.g. `chi_squared`, `ContinuousFunction::tMin`, or `CurveViewRange::xMin`). We should definitely move those to a single floating-point type (and I think this means `double`).
* Using `size_t` and `int` inconsistently (those aren't the same type)

### DIFF
--- a/apps/graph/graph/graph_controller.cpp
+++ b/apps/graph/graph/graph_controller.cpp
@@ -129,8 +129,8 @@ float GraphController::interestingXHalfRange() const {
     }
     // Compute the combined range of the functions
     assert(f->plotType() == ContinuousFunction::PlotType::Cartesian); // So that tMin tMax represents xMin xMax
-    tMin = std::min(tMin, f->tMin());
-    tMax = std::max(tMax, f->tMax());
+    tMin = std::min<double>(tMin, f->tMin());
+    tMax = std::max<double>(tMax, f->tMax());
   }
   constexpr float rangeMultiplicator = 1.6f;
   if (characteristicRange > 0.0f ) {
--- a/apps/graph/graph/graph_controller.cpp
+++ b/apps/graph/graph/graph_controller.cpp
@@ -1,15 +1,11 @@
 #include "graph_controller.h"
 #include "../app.h"
+#include <algorithm>
 
 using namespace Poincare;
 using namespace Shared;
 
 namespace Graph {
-
-static inline float minFloat(float x, float y) { return x < y ? x : y; }
-static inline float maxFloat(float x, float y) { return x > y ? x : y; }
-static inline double minDouble(double x, double y) { return x < y ? x : y; }
-static inline double maxDouble(double x, double y) { return x > y ? x : y; }
 
 GraphController::GraphController(Responder * parentResponder, ::InputEventHandlerDelegate * inputEventHandlerDelegate, Shared::InteractiveCurveViewRange * curveViewRange, CurveViewCursor * cursor, int * indexFunctionSelectedByCursor, uint32_t * modelVersion, uint32_t * previousModelsVersions, uint32_t * rangeVersion, Poincare::Preferences::AngleUnit * angleUnitVersion, ButtonRowController * header) :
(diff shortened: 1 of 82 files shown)


apps/graph/graph/graph_controller.cpp
```diff
@@ -145,7 +141,7 @@ float GraphController::interestingXHalfRange() const {
   if (tMin >= -defaultXHalfRange && tMax <= defaultXHalfRange) {
     /* If the combined Range of the functions is smaller than the default range,
      * use it. */
-    float f = rangeMultiplicator * (float)maxDouble(std::fabs(tMin), std::fabs(tMax));
+    float f = rangeMultiplicator * (float)std::max(std::fabs(tMin), std::fabs(tMax));
```
LeaNumworks, line 144: Remove (float)

apps/probability/distribution/chi_squared_distribution.cpp
```diff
@@ -79,7 +78,7 @@ double ChiSquaredDistribution::cumulativeDistributiveInverseForProbability(doubl
     2.0 * *probability * std::exp(std::lgamma(ceilKOver2)) / (exp(-kOver2Minus1) * std::pow(kOver2Minus1, kOver2Minus1)) :
     30.0; // Ad hoc value
   xmax = std::isnan(xmax) ? 1000000000.0 : xmax;
-  return cumulativeDistributiveInverseForProbabilityUsingIncreasingFunctionRoot(probability, FLT_EPSILON, maxDouble(xMax(), xmax));
+  return cumulativeDistributiveInverseForProbabilityUsingIncreasingFunctionRoot(probability, FLT_EPSILON, std::max(xMax(), xmax));
```
LeaNumworks: float/double comparison

apps/regression/graph_controller.cpp
```diff
@@ -390,8 +387,8 @@ InteractiveCurveViewRangeDelegate::Range GraphController::computeYRange(Interact
   for (int series = 0; series < Store::k_numberOfSeries; series++) {
     for (int k = 0; k < m_store->numberOfPairsOfSeries(series); k++) {
       if (m_store->xMin() <= m_store->get(series, 0, k) && m_store->get(series, 0, k) <= m_store->xMax()) {
-        minY = minFloat(minY, m_store->get(series, 1, k));
-        maxY = maxFloat(maxY, m_store->get(series, 1, k));
+        minY = std::min(minY, m_store->get(series, 1, k));
```
LeaNumworks: double/float comparison

apps/regression/store.cpp
```diff
@@ -199,15 +197,15 @@ void Store::resetMemoization() {
 float Store::maxValueOfColumn(int series, int i) const {
   float maxColumn = -FLT_MAX;
   for (int k = 0; k < numberOfPairsOfSeries(series); k++) {
-    maxColumn = maxFloat(maxColumn, m_data[series][i][k]);
+    maxColumn = std::max(maxColumn, m_data[series][i][k]);
```
LeaNumworks: float/double comparison

apps/regression/store.cpp
```diff
@@ -199,15 +197,15 @@ void Store::resetMemoization() {
...
   }
   return maxColumn;
 }
 
 float Store::minValueOfColumn(int series, int i) const {
   float minColumn = FLT_MAX;
   for (int k = 0; k < numberOfPairsOfSeries(series); k++) {
-    minColumn = minFloat(minColumn, m_data[series][i][k]);
+    minColumn = std::min(minColumn, m_data[series][i][k]);
```
LeaNumworks: float/double comparison

Issue #1506, opened 2020-04-12
turtle.circle() handles negative diameter


RedGl0w:
#### Describe the solution you'd like
I would like the turtle.circle() handles negative diameter, for drawing the circle backward.

RedGl0w:
Closed with https://github.com/numworks/epsilon/pull/1519

PR #1519, opened 2020-04-14
Negative diameter circle turtle support

RedGl0w:
This patch https://github.com/numworks/epsilon/issues/1506


artaxxx:
Thanks @RedGl0w ! It works great!

Ecco:
One quick remark : aren't we re-inventing `fabs()` here? We were already using the same construct though.

RedGl0w:
> fabs()

RedGl0w:
Sorry misscliclk :/

RedGl0w:
> One quick remark : aren't we re-inventing `fabs()` here? We were already using the same construct though.

If you are talking of
```cpp
(radius > 0 ? 1 : -1)
```
This was already used in the same line, so I thought the same it was, the better it was... If you prefer a cleaner solution, I can change the whole line :smile:

PR #1527, opened 2020-04-15
added german translation / changed some wrong ones

maximilianwalter:
Added a few German translations. But I didn't know if I could translate the Dutch Exam mode messages (and similar warnings) as well, so I didn't do that for now.
I will probably translate the whole Python stuff (apps/code/catalog.de.i18n) but that will take a bit longer. So for now just this.


EmilieNumworks:
Thanks @maximilianwalter!

PR #1528, opened 2020-04-19
Port Epsilon to the Nintendo 3DS

M4xi1m3:
This is a fully functional port of Epsilon to the Nintendo 3DS. It has display, keyboard and button support. It runs well on citra and on real hardware.

**Building:**
You need devkitPro and devkitARM installed and in your path (instructions [here](https://devkitpro.org/wiki/Getting_Started))
```
make PLATFORM=simulator TARGET=3ds -j
3dsxtool output/release/simulator/3ds/epsilon.elf epsilon.3dsx
```
You can then put epsilon.3dsx on a SD card to run it from the HBC or use 3dslink to launch it over the network:
```
3dslink epsilon.3dsx -a <3DS' IP ADDRESS>
```

**Todo list:**
- [x] Get it to boot
- [x] Get it to display stuff
- [x] Virtual keyboard
  - [ ] Put the non-debug keyboard image
  - [ ] Add a debouncer for the virtual keyboard
- [x] Button support
  - [ ] Add a debouncer for the buttons
- [x] Add language detection
- [ ] Add a way to quit properly (without force-shutting down the system)
- [ ] LED Driver
- [ ] Edit build system to make it output a .3dsx
  - [ ] Make it integrate sdmh data
- [ ] Edit the build system to make it output a .cia
  - [ ] Make the .cia able to overwrite the home menu
- [ ] Go to an exam with that thing

**Disclaimer:**
This is obviously a joke, but at the same time a fun PoC.

**Screenshots:**
Mandelbrot example + function drawing on citra:
Peek 19-04-2020 14-31
Picture of it running on actual hardware:
3ds


Ecco:
LOL 🤣

By the way, I'm surprised that you would need a debouncer. Did you actually experience any bounce in practice on real hardware?

That being said, and no matter how fun this is, I'll have to close this PR. We don't have any plan of merging nor maintaining this port (for reasons I assume will be obvious to everyone). And we'd rather keep only active PRs open.

Thanks though, that's a lot of fun 😄

M4xi1m3:
Thank you for your kind answer. I'm planning to continue working on this, for the meme, even tho it won't be merged. I actually experienced bounciness on real hardware, both with the buttons and the touchscreen. I don't know if it's because my 3DS is getting old or anything, but I prefer to put a debouncer just in case. I might put a build of this on the TinyDB 3DS homebrew repository once it's mature enough. I also found a weird bug involving pressing shift / alpha. It only stays activated for like one second, even if I don't do anything.

Issue #1533, opened 2020-04-23
Proper handling of colors in Python


Ecco:
#### Problem you'd like to fix

Specifying a color in Python is lacking:
* We don't support named colors
* We don't support float-based ranges
* It's been re-implemented differently in `mod_kandinsky`, `mod_turtle`, and `mod_matplotlib`.

#### Describe the solution you'd like

Implement a unified `parseColor` function that lives in Python only (we specifically don't want to use those features outside of Python), and then use that function from the 3 aforementioned modules. The function should be defined in `python/port/port.(h/cpp)` with the following signature:

```c++
namespace MicroPython{

enum class ColorMode {
  MaxIntensity1,
  MaxIntensity255
};

KDColor MicroPython::ParseColor(mp_obj_t input, ColorMode mode);
}
```

It should be able to parse the following parameters:
|Type|Value|ColorMode|
|-|-|-|
|String| `"#FF0000"`|Any|
|String|`"#ff0000"`|Any|
|String|`0.5`|1|
|String| `"red"`|Any|
|String| `"r"` | Any |
|Tuple of int/float | `(1.0, 0.0, 0.0)` | 1 |
|Tuple of int/float| `(255, 0, 0)` | 255 |

#### Notes

* As far as we know, neither Turtle nor Matplotlib support passing a color from a single int (e.g. `0xFF0000`) or float (e.g. `0.5`) value, so we're better off not do this for now.
* The ColorMode parameter should always be 1 in `modmatplotlib`, always be 255 in `modkandinsky`, and should be configurable in `modturtle` (through a `colormode` function) with a default value of 1.

RedGl0w:
I don't exactly understand what's the `0.5` string is
By the way, you could also support int as colors, because if you add `"#ffffff` why not adding `0xffffff`?

Ecco:
> I don't exactly understand what's the `0.5` string is

It's a mid-grey. See [matplotlib's spec](https://matplotlib.org/tutorials/colors/colors.html).

> By the way, you could also support int as colors, because if you add `"#ffffff` why not adding `0xffffff`?

See the note above 😄 It's not supported elsewhere so let's not bother for now 😄

RedGl0w:
> ```c++
> namespace MicroPython{
> 
> enum class ColorMode {
>   MaxIntensity1,
>   MaxIntensity255
> };
> ```

[...]

PR #1534, opened 2020-04-25
[Python] universal colors

RedGl0w:
This partially fixes https://github.com/numworks/epsilon/issues/1533 and fixes https://github.com/numworks/epsilon/issues/1489
**Todo :**

- [x] Add `colormode()` into the toolbox
- [ ] Move colors to a new category in toolbox
- [ ] Add the colors in pyplot (In a future PR)


python/src/py/obj.h
```diff
@@ -90,6 +90,7 @@ static inline bool mp_obj_is_qstr(mp_const_obj_t o)
     { return ((((mp_int_t)(o)) & 3) == 2); }
 #define MP_OBJ_QSTR_VALUE(o) (((mp_uint_t)(o)) >> 2)
 #define MP_OBJ_NEW_QSTR(qst) ((mp_obj_t)((((mp_uint_t)(qst)) << 2) | 2))
+mp_obj_t mp_obj_new_int_via_str(const char* value, int base);
```
EmilieNumworks, line 93: We try to avoid changing the external code. The content of python/src/py is identical to the MicroPython repository. This code should go somewhere in python/src/port/.

RedGl0w: I thought this was the better place :/ I will try to move it in port.cpp

python/src/py/objfloat.c
```diff
@@ -201,6 +201,24 @@ mp_obj_t mp_obj_new_float(mp_float_t value) {
     return MP_OBJ_FROM_PTR(o);
 }
 
+mp_obj_t mp_obj_new_float_via_str(const char* value) {
```
EmilieNumworks, line 204: Why don't you create a mp_float_t directly instead of allocing extra mp_obj_t on the Python heap?

python/port/port.cpp
```diff
@@ -170,6 +172,83 @@ void MicroPython::collectRootsAtAddress(char * address, int byteLength) {
 #endif
 }
 
+KDColor MicroPython::ColorParser::ParseColor(mp_obj_t input, ColorModes ColorMode){
+  if(MP_OBJ_IS_STR(input)){
```
EmilieNumworks, line 176: Maybe use mp_obj_is_str for more consistency?

python/src/py/objfloat.c
```diff
@@ -201,6 +201,24 @@ mp_obj_t mp_obj_new_float(mp_float_t value) {
...
+    mp_obj_float_t *o = m_new(mp_obj_float_t, 1);
```
EmilieNumworks, line 205: I don't exactly get what you're doing.
Can't you use this code instead:
`mp_parse_num_decimal(value, strlen(value), false, false, NULL)`
?

RedGl0w: I firstly need to get the length of the string, until the firsT \X0

python/src/py/objint.c
```diff
@@ -462,3 +462,17 @@ const mp_obj_type_t mp_type_int = {
     .binary_op = mp_obj_int_binary_op,
     .locals_dict = (mp_obj_dict_t*)&int_locals_dict,
 };
+
+mp_obj_t mp_obj_new_int_via_str(const char* value, int base) {
```
EmilieNumworks, line 466: Same comment as mp_obj_new_float_via_str:
could it be replaced by `mp_parse_num_integer(value, strlen(value), 16, NULL)`?

RedGl0w: Like I said, I firstly need to get the length of the string until the first \x0

python/port/mod/turtle/modturtle.cpp
```diff
@@ -131,37 +133,63 @@ mp_obj_t modturtle_isdown() {
...
 mp_obj_t modturtle_pencolor(size_t n_args, const mp_obj_t *args) {
-  if (n_args == 0) {
-    // pencolor()
-    KDColor c = sTurtle.color();
-    mp_obj_t mp_col[3];
-    mp_col[0] = mp_obj_new_int_from_uint(c.red());
-    mp_col[1] = mp_obj_new_int_from_uint(c.green());
-    mp_col[2] = mp_obj_new_int_from_uint(c.blue());
-    return mp_obj_new_tuple(3, mp_col);
+  switch(n_args){
```
EmilieNumworks, line 136: Can't we factorize the construction of the returned tuple in case 0 and case 3?

python/port/mod/turtle/modturtle.cpp
```diff
@@ -131,37 +133,63 @@ mp_obj_t modturtle_isdown() {
...
+        mp_col[2] = mp_obj_new_float(c.blue() / 255.0);
+      }
+      return mp_obj_new_tuple(3, mp_col);
+    }
+    case 1:{
+      sTurtle.setColor(MicroPython::ColorParser::ParseColor(args[0], sTurtle.colorMode()));
+      break;
+    }
+    case 3:{
+      // pencolor(r, g, b)
```
EmilieNumworks, line 157: The following code is a duplicate of a part of `MicroPython::ColorParser::ParseColor`. We should create a tuple and call with a helper `ColorTuple(args[0], args[1],args[2])` (which takes the mode into account) and then call `MicroPython::ColorParser::ParseColor`.

python/port/mod/turtle/modturtle.cpp
```diff
@@ -131,37 +133,63 @@ mp_obj_t modturtle_isdown() {
...
-    mp_col[0] = mp_obj_new_int_from_uint(c.red());
-    mp_col[1] = mp_obj_new_int_from_uint(c.green());
-    mp_col[2] = mp_obj_new_int_from_uint(c.blue());
-    return mp_obj_new_tuple(3, mp_col);
+  switch(n_args){
+    case 0:{
+      // pencolor()
+      KDColor c = sTurtle.color();
+      mp_obj_t mp_col[3];
+      if(sTurtle.colorMode() == MicroPython::ColorParser::ColorModes::MaxIntensity255){
```
EmilieNumworks, line 141: You could create a helper `ColorTuple(args[0], args[1],args[2])` (in order to use it in case 3

python/port/port.h
```diff
@@ -39,6 +41,31 @@ void deinit();
...
+      m_color(color)
+    {}
+    const char * name() const { return m_name; }
+    KDColor color() const { return m_color; }
+  private:
+    const char * m_name;
+    KDColor m_color;
+  };
+
+  public:
```
EmilieNumworks, line 59: Idea:
```
enum class ColorModes {
   MaxIntensity1 = 1,
   MaxIntensity255 = 255,
   };
```
and replace magic number 255 by `(int)ColorModes::MaxIntensity255` ?

python/port/port.h
```diff
@@ -39,6 +41,31 @@ void deinit();
...
+  enum class ColorModes {
```
EmilieNumworks, line 60: Singular `ColorMode` as class names which are usually singular

python/src/py/objfloat.c
EmilieNumworks: `strlen` would do that, don't it?

EmilieNumworks:
Nice PR! Can I commit on your branch in order to do the few changes I suggested and to take care of pyplot?

EmilieNumworks:
@artaxxx, what do you think about the toolbox? This PR adds the enhancement asked in #1533. Should we add the `colormode` in the turtle module?

RedGl0w:
> Nice PR! Can I commit on your branch in order to do the few changes I suggested and to take care of pyplot?

You can, of course !

artaxxx:
Sure, `colormode()` should be added to Modules > Turtle between `color` and `'blue'` as well as in the Catalog!

We should add it as:
`colormode(x)`
`Set the color mode to 1.0 or 255` in EN
`Met le mode de couleur à 1.0 ou 255` in FR if it fits

RedGl0w:
> Sure, `colormode()` should be added to Modules > Turtle between `color` and `'blue'` as well as in the Catalog!
> 
> We should add it as:
> `colormode(x)`
> `Set the color mode to 1.0 or 255` in EN
> `Met le mode de couleur à 1.0 ou 255` in FR if it fits

But I think the colors should go in a new special category, because they will be the same in kandinsky, turtle and pyplot, so it would make a lot of duplicates :/

python/port/port.h
Ecco: Even better : use a static_cast instead of a C-style cast 😄

EmilieNumworks:
The color handling on pyplot is still missing but I'll do it in another PR as it requires some prior work on the pyplot module.

RedGl0w:
> The color handling on pyplot is still missing but I'll do it in another PR as it requires some prior work on the pyplot module.

That was I was thinking. Now we just need to move to a new category the colors in toolbox.

EmilieNumworks:
The module menu of the toolbox is made to easily expose all methods of a module to the user. Duplicates are not an issue here if it's made easier for the end user.

apps/code/catalog.universal.i18n
```diff
@@ -208,6 +208,8 @@ PythonTurtleCommandBrown = "'brown'"
 PythonTurtleCommandCircle = "circle(r)"
 PythonTurtleCommandColor = "color('c')/color(r,g,b)"
 PythonTurtleCommandColorWithoutArg = "color(\x11)"
+PythonTurtleCommandColorMode = "colormode(x)"
+PythonTurtleCommandColorModeWithoutArg = "colormode(\x11)"
```
LeaNumworks, line 212: No need to define PythonTurtleCommandColorModeWithoutArg!
It should be removed from PythonTurtleCommandColorMode by our algorithm

apps/code/python_toolbox.cpp
```diff
@@ -145,6 +145,7 @@ const ToolboxMessageTree TurtleModuleChildren[] = {
   ToolboxMessageTree::Leaf(I18n::Message::PythonTurtleCommandShowturtle, I18n::Message::PythonTurtleShowturtle, false),
   ToolboxMessageTree::Leaf(I18n::Message::PythonTurtleCommandHideturtle, I18n::Message::PythonTurtleHideturtle, false),
   ToolboxMessageTree::Leaf(I18n::Message::PythonTurtleCommandColor, I18n::Message::PythonTurtleColor, false, I18n::Message::PythonTurtleCommandColorWithoutArg),
+  ToolboxMessageTree::Leaf(I18n::Message::PythonTurtleCommandColorMode, I18n::Message::PythonTurtleColorMode, false, I18n::Message::PythonTurtleCommandColorModeWithoutArg),
```
LeaNumworks, line 148: Remove PythonTurtleCommandColorModeWithoutArg

LeaNumworks: And remove the `false` to use the algorithm to remove the argument  "x" from `I18n::Message::PythonTurtleColorMode.`

It should thus be: `ToolboxMessageTree::Leaf(I18n::Message::PythonTurtleCommandColorMode, I18n::Message::PythonTurtleColorMode)`

python/port/mod/kandinsky/modkandinsky.cpp
```diff
@@ -1,32 +1,16 @@
...
-
-    return KDColor::RGB888(
-      mp_obj_get_int(elem[0]),
-      mp_obj_get_int(elem[1]),
-      mp_obj_get_int(elem[2])
-    );
-}
-
-static mp_obj_t TupleForRGB(uint8_t r, uint8_t g, uint8_t b) {
+static mp_obj_t TupleForRGB(KDColor c) {
```
LeaNumworks, line 9: Rename? For instance TupleForKDColor?

Issue #1529, opened 2020-04-21
Add y = ax model in the Regression app


artaxxx:
#### Problem you'd like to fix
This model is not available and could be very useful! 

#### Screenshots
![screenshot (71)](https://user-images.githubusercontent.com/20040032/79853035-ced4bb00-83c7-11ea-99cf-ffd8196a9c08.png)

BoltzmannSolver:
It could be useful to add R² parameter in proportional regression.

LeaNumworks:
> It could be useful to add R² parameter in proportional regression.

Hello @BoltzmannSolver, thanks for the suggestion!
We plan to add R² for all regressions, see #1530

PR #1537, opened 2020-04-29
[Regression] Add linear regression

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/1529
Todo :
- [ ] Add translations
- [x] Add the display of r and r^2 in the calculation controller for affine regression


adriweb:
Hm. Linear is supposed to be `a*x` and affine `a*x+b`.
Looks like it's reversed here, and the real affine is just already implemented under the name linear.

RedGl0w:
> Hm. Linear is supposed to be `a*x` and affine `a*x+b`.
> Looks like it's reversed here, and the real affine is just already implemented under the name linear.

Oh fu** I haven't realized when I did that :sob:
I hope my math teacher will never see that :troll:
I'll patch that in a few minutes !
Thank you !

apps/regression/base.universal.i18n
```diff
@@ -1,3 +1,4 @@
+AffineRegressionFormula  = " y=a·x "
```
LeaNumworks, line 1: This is the LinearRegressionFormula!

`LinearRegressionFormula  = " y=a·x+b "` is defined in `apps/shared.universal.i18n` but should be renamed AffineRegressionFormula there and in `apps/graph/graph/banner_view.cpp`, and then used in the affine model.

apps/regression/model/affine_model.cpp
```diff
@@ -0,0 +1,52 @@
+#include "linear_model.h"
```
LeaNumworks, line 1: Should include affine_model.h

apps/regression/model/affine_model.h
```diff
@@ -0,0 +1,24 @@
+#ifndef REGRESSION_LINEAR_MODEL_H
```
LeaNumworks, line 1: The header guardian should be REGRESSION_AFFINE_MODEL_H

apps/regression/base.universal.i18n
RedGl0w: Yes the switch between affine and linear was a bit chaotic

apps/regression/model/linear_model.cpp
```diff
@@ -1,4 +1,4 @@
-#include "linear_model.h"
+#include "affine_model.h"
```
LeaNumworks, line 1: linear_model.cpp should include linear_model.h

apps/regression/model/linear_model.cpp
```diff
@@ -28,7 +27,7 @@ double LinearModel::levelSet(double * modelCoefficients, double xMin, double ste
   if (a == 0) {
     return NAN;
   }
-  return (y-b)/a;
+  return y-b;
```
LeaNumworks, line 30: There is no b coefficient in the linear model, so the line `double b = modelCoefficients[1];` should be removed.
The levelSet is supposed to give x given y, so here it should return y/a

apps/regression/model/linear_model.cpp
```diff
@@ -28,7 +27,7 @@ double LinearModel::levelSet(double * modelCoefficients, double xMin, double ste
   if (a == 0) {
```
LeaNumworks, line 27: As you are modifying this file anyways, could you please change this to `a == 0.0` ? Thanks!

apps/regression/model/linear_model.cpp
LeaNumworks: A few lines below, in fit, please remove `modelCoefficients[1] = store->yIntercept(series);`

LeaNumworks: And again in `LinearModel::partialDerivate`, the case
` if (derivateCoefficientIndex == 1) {
    // Derivate: 1;
    return 1;
  }`
should not exist

apps/regression/model/linear_model.h
```diff
@@ -1,5 +1,5 @@
-#ifndef REGRESSION_LINEAR_MODEL_H
-#define REGRESSION_LINEAR_MODEL_H
+#ifndef REGRESSION_AFFINE_MODEL_H
```
LeaNumworks, line 1: The header guardians should match the file name, so REGRESSION_LINEAR_MODEL_H here

LeaNumworks:
Hi, thanks for the changes!

LeaNumworks:
I took the liberty to rebase your PR, so I push-forced it.

We decided to keep the linear regression as aX+b and to rename aX as Proportional regression. 
I also added commits that do small cleanings and add a test.
I removed r and r2, we will do it later.

Thanks again!

Issue #666, opened 2018-09-23
Setting the table's interval doesn't work without changing parameters


FramboisePi:
- Open Function app
- Enter a function like x^2 and go to Table tab
- Make some manual change to the table, like deleting the firsts lines
- Open "Set the Interval" and choose "Confirm" without changing the start, end and step

The table doesn't change.
It should use x values from 1 to 10.

It works fine if we change a parameter.
This behaviour is present on Numworks and on the simulator

![function_-set-the-interval](https://user-images.githubusercontent.com/25133304/45926729-537b0100-bf27-11e8-849e-0e017ea388c8.gif)

PR #1538, opened 2020-05-01
[Interval] Patch the need recomputing

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/666

### DIFF
--- a/apps/shared/interval.h
+++ b/apps/shared/interval.h
@@ -18,7 +18,6 @@ public:
   void setEnd(double f);
   void setStep(double f);
   void setElement(int i, double f);
-  bool needCompute() const { return m_needCompute; }
   void setNeedCompute(bool compute){ m_needCompute = compute;}
   void reset();
   void clear();
--- a/apps/shared/interval.h
+++ b/apps/shared/interval.h
@@ -18,6 +18,8 @@ public:
   void setEnd(double f);
   void setStep(double f);
   void setElement(int i, double f);
+  bool NeedRecompute() const { return m_needCompute; }
+  void setNeedRecompute(bool needrecompute);
   void reset();
   void clear();
   // TODO: decide the max number of elements after optimization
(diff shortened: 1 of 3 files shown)


Ecco:
Hi @RedGl0w ! Thanks for this :) I don't know if this is the right fix, but it's so short I would assume it is. On the PR itself, just a few notes:
* Please put the setter directly in the header file.
* The getter should start with a lowercase letter
* It'd be better if the setter/getter matched the instance variable's name. So `bool needCompute()` and `void setNeedCompute(bool)`

RedGl0w:
>     * The getter should start with a lowercase letter

It seems I won't ever manage to respect camelCase :sob:

apps/shared/interval.h
```diff
@@ -18,6 +18,8 @@ class Interval {
   void setEnd(double f);
   void setStep(double f);
   void setElement(int i, double f);
+  bool needCompute() const { return m_needCompute; }
```
EmilieNumworks, line 21: This does't need to be exposed, does it?

RedGl0w: We don't need it, you're right.
I always add setter and getter, even when they're useless

apps/shared/interval.h
```diff
@@ -18,6 +18,7 @@ class Interval {
   void setEnd(double f);
   void setStep(double f);
   void setElement(int i, double f);
+  void setNeedCompute(bool compute){ m_needCompute = compute;}
```
EmilieNumworks, line 21: m_needCompute is never set to false through external code, only Interval::computeElements private method will turn m_needCompute to false. So the public API of Interval should only expose a method that can turn m_needCompute to true like `forceComputing()` - you might think of a better name!

RedGl0w: Ok !

EmilieNumworks:
Thanks @RedGl0w, this fixes an old bug!

PR #1544, opened 2020-05-06
[escher] View: remove useless virtual destructor

EmilieNumworks:


escher/include/escher/view.h
```diff
@@ -43,6 +51,7 @@ class View {
   View(const View& other) = delete;
```
EmilieNumworks, line 51: Leave copy/move assignment as deleted

PR #1548, opened 2020-05-11
[App/zoom_parameter_controller] Changed subview order

### LINKED ISSUE #500: [Functions] Minor display glitch when waking the device up from sleep
This is a minor and very specific bug.

Just updated with today version. Don't know if this bug also exist on older version.

1) Go on fonction and create a f(x) = sin(x) fonction.
2) Show the graphical representation of the function, then open the zoom menu
3) Turn of the numworks, then turn it on again.
4) For a small duration you should see a glitch at the bottom (where zoom key are written).

This bug also happens with cos, square and frac function, but doesn't happens with all function.
With the floor and ceil function, the glitch is only on the bottom right section, but the size of the glitch vary with zoom and position.
For square root and ln it happen if you move on the right.

This is probably related to the time required to draw the function.

RedGl0w:
This fixes #500
And I can't find any consequences of this switch.

### DIFF
--- a/apps/shared/zoom_parameter_controller.cpp
+++ b/apps/shared/zoom_parameter_controller.cpp
@@ -58,6 +58,11 @@ int ZoomParameterController::ContentView::numberOfSubviews() const {
 
 View * ZoomParameterController::ContentView::subviewAtIndex(int index) {
   assert(index >= 0 && index < 2);
+  /* The order of subview is important here :
+   * If we redraw the curveView before the legendView, that can have some display issue, when exiting sleep mode, which
+     can be visible, if the redraw of curveView is long (with complicated curve), so we prefer to have legendView
+    at first subview.
+  */
   if (index == 0) {
     return &m_legendView;
   }
--- a/apps/shared/zoom_parameter_controller.cpp
+++ b/apps/shared/zoom_parameter_controller.cpp
@@ -59,9 +59,9 @@ int ZoomParameterController::ContentView::numberOfSubviews() const {
 View * ZoomParameterController::ContentView::subviewAtIndex(int index) {
   assert(index >= 0 && index < 2);
   if (index == 0) {
-    return m_curveView;
+    return &m_legendView;
   }
-  return &m_legendView;
+  return m_curveView;
 }
 
 void ZoomParameterController::ContentView::layoutSubviews(bool force) {


apps/shared/zoom_parameter_controller.cpp
```diff
@@ -59,9 +59,9 @@ int ZoomParameterController::ContentView::numberOfSubviews() const {
 View * ZoomParameterController::ContentView::subviewAtIndex(int index) {
   assert(index >= 0 && index < 2);
   if (index == 0) {
-    return m_curveView;
+    return &m_legendView;
```
EmilieNumworks, line 62: This is indeed a good trick to fix the bug. A comment explaining why this order (because, now, it matters) would be nice!

RedGl0w: The subview are redrawn one after the other, so by changing the subview order, the legendview will be redraw before the curveview, and the legendview, which wasn't redrawn, so shown like a sort of glitch isn't anymore.

EmilieNumworks: I meant in the code :)
In the same way it is done in apps/calculation/history_view_controller.cpp (only the first part of the comment):
```
View * HistoryViewCell::subviewAtIndex(int index) {
  /* The order of the subviews should not matter here as they don't overlap.
   * However, the order determines the order of redrawing as well. For several
   * reasons listed after, changing subview selection often redraws the entire
   * m_scrollableOutputView even if it seems unecessary:
   * - Before feeding new Layouts to ExpressionViews, we reset the hold layouts
   *   in order to empty the Poincare pool and have more space to compute new
   *   layouts.
   * - Even if we did not do that, ExpressionView::setLayout doesn't avoid
   *   redrawing when the previous expression is identical (for reasons
   *   explained in expression_view.cpp)
   * - Because of the toggling burger view, ExpressionViews often have the same
   *   absolute frame but a different relative frame which leads to redrawing
   *   them anyway.
   * All these reasons cause a blinking which can be avoided if we redraw the
   * output view before the input view (starting with redrawing the more
   * complex view enables to redraw it before the vblank thereby preventing
   * blinking).
   * TODO: this is a dirty hack which should be fixed! */
  View * views[3] = {&m_scrollableOutputView, &m_inputView, &m_ellipsis};
  return views[index];
}
```

RedGl0w: done :+1:

PR #1554, opened 2020-05-15
Python garbage collection

EmilieNumworks:
The Python garbage collection failed to mark some allocated objects for x86-64 architecture. We used to rely on the setjmp to get the registers values but this was not true for all architectures.

### DIFF
--- a/ion/src/simulator/shared/collect_registers.cpp
+++ b/ion/src/simulator/shared/collect_registers.cpp
@@ -5,10 +5,12 @@ extern "C" {
 // define in assembly code
 // Force the name as archs (linux/macos) don't mangle C names the same way
 extern uintptr_t collect_registers(uintptr_t * regs) asm ("_collect_registers");
+
 }
 namespace Ion {
 
-// Wrapper to avoid
+// Wrapper to avoid handling c++ name mangling when writing assembly code
+
 uintptr_t collectRegisters(jmp_buf buf) {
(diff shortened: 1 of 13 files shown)


python/port/port.cpp
```diff
@@ -155,29 +155,12 @@ void MicroPython::registerScriptProvider(ScriptProvider * s) {
 }
 
 void MicroPython::collectRootsAtAddress(char * address, int byteLength) {
-#if __EMSCRIPTEN__
-   // All objects are aligned, as asserted.
+  /* All pointers on the stack are aligned on sizeof(void *), as asserted.
+   * This is a consequence of the alignment requireduirements of compilers as long
```
boricj, line 159: Typo: "requireduirements". Comment overall seems corrupted or truncated.

ion/src/simulator/shared/collect_registers.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+extern "C" {
+
+// define in assembly code
+// Force the name as archs (linux/macos) don't mangle C names the same way
+extern uintptr_t collect_registers(uintptr_t * regs) asm ("_collect_registers");
+}
+namespace Ion {
+
+// Wrapper to avoid
```
boricj, line 11: "Wrapper to avoid"... ?

EmilieNumworks: Oups

PR #1556, opened 2020-05-18
Send file in emscripten

RedGl0w:
This PR adds a button for sending file in web simulator.

**Todo / fixme** :
- [x] Add multiple file sending
- [x] Patch issue with file content > 3*sizeof(char)
- [x] Change button style (I'm not not enough good for doing logo : 
image)
- [x] Change error message display (it's still ugly, but more userfriendly)
- [x] Add download
- [x] Add download interface (which is ugly)
- [x] Add script name upload regex
- [ ] Patch issue when uploading multiple file ~~(it seems that `strlen` has an issue, and is reading the beginning of the next record :| )~~ (nope)

### DIFF
--- a/ion/src/shared/storage.cpp
+++ b/ion/src/shared/storage.cpp
@@ -5,7 +5,6 @@
 #if ION_STORAGE_LOG
 #include<iostream>
 #endif
-
 namespace Ion {
 
 /* We want to implement a simple singleton pattern, to make sure the storage is
@@ -597,7 +596,7 @@ Storage::RecordIterator & Storage::RecordIterator::operator++() {
 
 #if EPSILON_SDL_SCREEN_ONLY
 //FIXME Add errors support
-void IonStorageAddScript(const char* name, const char* content){
-  Ion::Storage::sharedStorage()->createRecordWithExtension(name, "py", content, sizeof(content));
+int IonStorageAddScript(const char* name, const char* content){
+  return static_cast<int>(Ion::Storage::sharedStorage()->createRecordWithExtension(name, "py", content, strlen(content)));
 }
 #endif
--- a/ion/src/shared/storage.cpp
+++ b/ion/src/shared/storage.cpp
@@ -595,7 +595,6 @@ Storage::RecordIterator & Storage::RecordIterator::operator++() {
 }
 
 #if EPSILON_SDL_SCREEN_ONLY
-//FIXME Add errors support
 int IonStorageAddScript(const char* name, const char* content){
   return static_cast<int>(Ion::Storage::sharedStorage()->createRecordWithExtension(name, "py", content, strlen(content)));
 }
--- a/ion/src/shared/storage.cpp
+++ b/ion/src/shared/storage.cpp
@@ -5,6 +5,7 @@
 #if ION_STORAGE_LOG
 #include<iostream>
 #endif
+#include <apps/code/script_store.h>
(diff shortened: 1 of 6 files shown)


cent20net:
Excellent improvement. 
Being able to load a python file directly on the emulator is a significant improvement

cent20net:
It works on https://epsilon.nsi.xyz/ 
Thanks RedGl0w, I hope NumWorks will accept this PR.

Ecco:
I'm not sure how it would fit in our current offering. The [online simulator](https://www.numworks.com/simulator/) is targeted at people using the entire calculator, and the [Python Workshop](https://workshop.numworks.com/python) for coding.

cent20net:
> I'm not sure how it would fit in our current offering. The [online simulator](https://www.numworks.com/simulator/) is targeted at people using the entire calculator, and the [Python Workshop](https://workshop.numworks.com/python) for coding.

FR
Bonjour Ecco,
Je demande un jour à un élève de me préparer un script en python pour la numworks, en classe 2nde. 
L'élève sérieux et motivé m'amène une clée USB avec le script au format .py, ce qui en soit est logique.
J'ouvre le simulateur en ligne, pour vidéo-projeter le code et le résultat, et là impossible de faire un import mais surtout impossible de faire un copié collé du script dans le simulateur. 
On s'est retrouvé très con à devoir retaper tout le script, ... que de temps perdu ...
Alors oui on aurait pu ouvrir le workshop, mais pour y saisir un script il faut s'identifier ce qui n'est pas pratique quand on utilise, comme tous devrait le faire, un mot de passe par site internet.
Donc, une méthode alternative serait de permettre le copier coller et d'ouvrir la possibilité d'importer un script par un un simple CTRL+V
Autre option, une version sans identification du workshop, un bac à sable car dans le workshop on peut faire des copié collé.

EN
Hello Ecco,
One day I asked a 10th grade student to write a python script for the Numworks. The student, being serious and motivated, brang me an USB stick with the script as a .py file, which is logic.
I opened the online simulator, to videoproject the code and the result. It was impossible for me to import the script or to copy paste it into the simu.
We ended up losing a lot of time typing the script by hand.
Yes, we could have opened the workshop, but you need to log in. This isn't very practical when you have a password per site, as everyone should. 
An alternative method would be to allow copy-paste in the simulator, or to have a version of the workshop without authentication, like a sandbox, because we can copy-paste in the workshop.

Ecco:
That's super interesting! Thanks @cent20net ! Like you said, we already cover all this, and in theory the pupil would only have needed to share an URL. I'm curious why this wasn't the case by the way.

That being said, we could imagine adding such a feature to our Workshop. Thanks for the input!

ion/src/shared/storage.cpp
```diff
@@ -594,3 +597,26 @@ Storage::RecordIterator & Storage::RecordIterator::operator++() {
 }
 
 }
+
+#if EPSILON_SDL_SCREEN_ONLY
+int IonStorageAddScript(const char* name, const char* content){
+  if(Code::Script::nameCompliant(name)){
+    return static_cast<int>(Ion::Storage::sharedStorage()->createRecordWithFullName(name, content, strlen(content)));
```
RedGl0w, line 604: The whole window should be redrawn here

RedGl0w:
With epsilon 15 and the shared clipboard this is way less usefull

PR #1563, opened 2020-05-27
[Apps/Shared] Fixed interval and curve view exit

### LINKED ISSUE #1539: Setting interval without confirming it shouldn't recompute
#### Describe the bug
When in the graph app, in the list view, you set the interval, but you go back without confirming it, it will still update the intervaL

#### Screenshots
Peek 02-05-2020 08-57

#### To Reproduce
1. Go to graph app
2. Create a function
3. Go to the list view
4. Change the interval **without** confirming it
5. Press back

#### Expected behavior
The interval shouldn't change.

#### Environment
 - Epsilon master (9b419acf4039f2730d9763cf1831a24613b2797a)
 - The platform(s) on which the problem happens: all

RedGl0w:
That fixes #1539 and https://github.com/numworks/epsilon/issues/1539#issuecomment-633944257.

I think there's a lot of warnings generated, I'll try to patch them.


RedGl0w:
Request : like in text field, use a shared draft buffer, for avoiding this huge PR

apps/shared/interval_parameter_controller.cpp
```diff
@@ -16,6 +17,11 @@ IntervalParameterController::IntervalParameterController(Responder * parentRespo
   }
 }
 
+void IntervalParameterController::setInterval(Interval * interval) { 
+  m_interval = interval;
+  m_tempInterval = *interval;
```
HugoNumworks, line 22: There are heavy members in m_tempInterval that we do not need to keep track of ( m_intervalBuffer ).
It would be lighter to only keep track of m_start, m_end and m_step.
It could be achieved by storing them in a nested class in Interval (such as Interval::IntervalParameters) , and manipulating this class to store the temporary states.

apps/shared/range_parameter_controller.cpp
```diff
@@ -70,9 +71,14 @@ bool RangeParameterController::textFieldDidFinishEditing(TextField * textField,
   return false;
 }
 
+void RangeParameterController::setRange(InteractiveCurveViewRange * range){
+  m_interactiveRange = range;
+  m_tempInteractiveRange = *range;
```
HugoNumworks, line 76: Same could be achieved here to avoid keeping track of useless InteractiveCurveViewRange members.

apps/shared/interval_parameter_controller.cpp
```diff
@@ -5,6 +5,7 @@ namespace Shared {
 IntervalParameterController::IntervalParameterController(Responder * parentResponder, InputEventHandlerDelegate * inputEventHandlerDelegate) :
   FloatParameterController<double>(parentResponder),
   m_interval(nullptr),
+  m_tempInterval(),
```
HugoNumworks, line 8: Switching line 8 and 7 to avoid compilation warnings as you declared them in the other order in the header file.

apps/shared/interactive_curve_view_controller.cpp
```diff
@@ -13,10 +13,12 @@ InteractiveCurveViewController::InteractiveCurveViewController(Responder * paren
   m_modelVersion(modelVersion),
   m_previousModelsVersions(previousModelsVersions),
   m_rangeVersion(rangeVersion),
+  m_interactiveRange(interactiveRange),
```
HugoNumworks, line 16: m_interactRange should be initialized right before m_rangeButton, as it is declared in the header file, to avoid compilation warnings.

RedGl0w:
Oh, I think I *once again* did something wrong with git :sob:. Going to reopen it (sorry :( )

PR #1571, opened 2020-05-28
[apps/shared] Fixed a small mistake in parameter_text_field_delegate.cpp

### LINKED ISSUE #1570: Every parameter text field, when pressing backspace, doesn't clear, but add what's after the text in ram
Reported by `7 doritos` `#9136` on omega discord

#### Describe the bug
Scenario :
- go to solver app
- input this equation : `(5-x)e^(-0.2x)=0`
- go to the interval controller (when pressing solve)
- without editing a cell, press backspace on one of the cell
- see the equation you entered being added

#### Screenshots
Peek 28-05-2020 15-38

#### Expected behavior
I would expect the input to be cleared, and the editing mode activated (which it is).

#### Environment
 - Epsilon master
 - All

RedGl0w:
In this commit, there was a small oversight in https://github.com/numworks/epsilon/commit/7a4ee746b2a49a090cf920cfe53935ccdea7d607, which caused https://github.com/numworks/epsilon/issues/1570

So this PR fixes #1570

### DIFF
--- a/apps/shared/parameter_text_field_delegate.cpp
+++ b/apps/shared/parameter_text_field_delegate.cpp
@@ -5,6 +5,7 @@ namespace Shared {
 
 bool ParameterTextFieldDelegate::textFieldDidReceiveEvent(TextField * textField, Ion::Events::Event event) {
   if (event == Ion::Events::Backspace && !textField->isEditing()) {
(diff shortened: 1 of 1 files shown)


EmilieNumworks:
Yes, indeed. Thanks!

PR #1588, opened 2020-06-06
Added % in poincare

RedGl0w:
image

### DIFF
--- a/poincare/Makefile
+++ b/poincare/Makefile
@@ -102,6 +102,7 @@ poincare_src += $(addprefix poincare/src/,\
   matrix_trace.cpp \
   matrix_transpose.cpp \
   multiplication.cpp \
+  modulo.cpp \
   n_ary_expression.cpp \
   naperian_logarithm.cpp \
   norm_cdf.cpp \
--- a/poincare/include/poincare/expression.h
+++ b/poincare/include/poincare/expression.h
@@ -118,6 +118,8 @@ class Expression : public TreeHandle {
   friend class StoreNode;
   friend class SymbolNode;
 
+  friend class Modulo;
+
 public:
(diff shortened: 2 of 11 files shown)


Ecco:
I'm not a 100% sure we want to map `%` to the modulo operation. A lot of people use this symbol for actual percentages, and we've had request (even though not very clear) about supporting these…

RedGl0w:
> 
> 
> I'm not a 100% sure we want to map `%` to the modulo operation. A lot of people use this symbol for actual percentages, and we've had request (even though not very clear) about supporting these…

Currently, my PR will output an error if someone does `50%` for exemple, so the main question to think is : will you add the percent as percent in the future or not. I don't think this is the better idea : in Python the percent is the modulo operator, and one of the main selling argument was the python support (and people want more compatibility between calculation / code), so I don't think creating this difference is a good idea...

A2drien:
I think it would be better to display `3` instead of `8-5[8/5] = 3`.
I don't think if you do 5×7 you want to see `7+7+7+7+7 = 35`. 😉

RedGl0w:
> I think it would be better to display `3` instead of `8-5[8/5] = 3`.
> I don't think if you do 5×7 you want to see `7+7+7+7+7 = 35`. wink

Yes, some people have already told me (ex : A2drien :trollface: )

artaxxx:
I don't think we should use the percentage sign. It's very confusing. Doesn't that duplicate the `rem` function by the way?

RedGl0w:
I didn't know there was the rem function. And I don't think personally that the % symbol is confusing for rem...

RedGl0w:
Indeed, I think that neither the rem function nor % are perfect solution for that, but rem is a bit better. Closing

PR #1589, opened 2020-06-06
Improved Building Instructions

PolarToCartesian:
Added:
+ Instructions for generating binpacks
+ Instructions to use WSL if msys2 doesn't work

(Based on personal experience as a new epsilon user/builder)


EmilieNumworks:
Hello @PolarToCartesian, thanks for your PR. But we'd rather keep the SDK more simple.
I'll close this PR because:
- we haven't had many issues with msys2 so far;
- the TOS for building the official device are not the same as the https://www.numworks.com/legal/tos/ (you should not use the ACCEPT_OFFICIAL_TOS flag);
- binpack target is meant to build multiple binaries in addition to those of Epsilon (I guess the command you want is ‘make epsilon.two_binaries’). But there are many different targets for different purposes and as said, we’d rather keep it simple.

PR #1590, opened 2020-06-06
[escher] RGB files compatibility

quentinguidee:
This little pull request adds support for RGB files without alpha channel.

_I don't know if it's very useful in Epsilon_, but we use this in Omega for the following reason:

When a user creates a custom theme, and creates non-transparent icons, some programs (such as Affinity Designer) do not export the alpha channel, which is logical, but it crashes the compilation of Epsilon. So we needed it from our side since the icons of Omega are being redesigned in Affinity Designer.


LeaNumworks:
Hi @quentinguidee, thanks for this PR!
It is indeed not very useful for now, but it does not complicate the code and it is in the inliner.c, so it is okay :)
Could you please rebase it on master so it can pass the integration checks?

M4xi1m3:
An update to emscripten broke your build system by removing EMTERPRETIFY. His PR is actually based on your latest master.

RedGl0w:
Issue opened : #1592

LeaNumworks:
Hi again, we should have fixed the emscripten bug with #1597, could you please try to rebase your PR on master? Thanks!

Issue #1592, opened 2020-06-11
Emscripten 1.39.17 removes emterpreter


RedGl0w:
Emscripten 1.39.17 was released a few days ago (https://github.com/emscripten-core/emsdk/commit/d371fb6ffc4720be6b02260300d82baa2b834327), and one of the changes of this version is that it removes the emterpreter support, which was needed by epsilon (since https://github.com/numworks/epsilon/commit/67f082f4e88ccef505f5648c0d09c081afb0ca8c). So, the web build need to be updated.

RedGl0w:
_fixed_ by #1597

PR #1593, opened 2020-06-11
Made the web simu build on emscripten-upstream.

M4xi1m3:
Since Emscripten 1.39.17, emterpreter support was removed. Seems like a good time to switch to WebASM.
It works flawlessly on latest firefox and chromuim.
The `-s SINGLE_FILE=1` emcc option makes it embed the .wasm file in the final .js, thus not changing the way the simulator is distributed, and making it able to run without a web server.
Using `-s ASYNCIFY` removes the need for `EMTERPRETIFY_ASYNC`.
I also updated the CI to use the latest emscripten-upstream toolchain.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -70,7 +70,7 @@ jobs:
     steps:
       - uses: numworks/setup-emscripten@v2
         with:
-          sdk: latest-fastcomp
+          sdk: latest-upstream
       - uses: actions/checkout@v2
       - run: make -j2 PLATFORM=simulator TARGET=web
       - run: make -j2 PLATFORM=simulator TARGET=web epsilon.official.zip
--- a/build/toolchain.emscripten.mak
+++ b/build/toolchain.emscripten.mak
@@ -3,120 +3,20 @@ CXX = emcc
 LD = emcc
 CPP = cpp
 
-EMSCRIPTEN_ASYNC_SYMBOLS = \
-SAFE_HEAP_LOAD \
-SAFE_HEAP_STORE \
-_IonDisplayForceRefresh\
-_IonEventsEmscriptenKeyDown \
-_IonEventsEmscriptenKeyUp \
-_IonEventsEmscriptenPushEvent \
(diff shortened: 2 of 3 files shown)


RedGl0w:
This fixes #1592.

EmilieNumworks:
Thanks @M4xi1m3! WebASM would indeed be a good idea in the future but it is still not supported for [~10% of users](https://caniuse.com/#feat=wasm). So we'll stick to js for now and update the CI to use an older version of Emscripten (see PR #1597). We keep in mind your PR for later!

adriweb:
Note that there is a way to have wasm fallback to asm.js with feature detection, it's an emscripten build flag :)

M4xi1m3:
Looking at plain raw numbers, asm.js (which is still a dependency of Emterpreter, used on epsilon) has [less compatibility](https://caniuse.com/#feat=asmjs) than [wasm](https://caniuse.com/#feat=wasm) does (76% vs 91%), making this PR revelant.

Furthermore, I have measured a 150% performance increase using wasm over Emterpreter. The overall size of the simu is bigger (4MB with Emterpreter, 9MB with wasm) BUT this is with wasm bundled as b64 encoded data in the js file (to be run without a web-server). By removing `SINGLE_FILE=1`, the total size is actually equivalent or lower than with Emterpreter (around 3.5MB), and WASM compresses way better tan Emterpreter bytecode (using http compression). This is very useful when using a web server (like on your website).

Ecco:
Actually, [asm.js](https://en.wikipedia.org/wiki/Asm.js) is a *strict subset* of JS. In other words, browser not having explicit asm.js support can still execute asm.js code… They just don't take advantage of potential speedups. Believe me, we would have noticed if 1 user out of 4 couldn't run our simulator 😄 

Regarding the execution speed, we don't care at all since it's already much faster than the actual device. Of course we'd like a smaller binary but our test didn't yield a difference significant enough to justify switching.

All this being said, we'll most likely eventually switch to WebAssembly, but given the current drawbacks it's not something we want to do just yet.

RedGl0w:
Note that it would also have closed #1607 and #1735

PR #1596, opened 2020-06-15
[docs] Add dependencies in windows SDK

EmilieNumworks:


Ecco:
Hmm, several questions:
 * We don't need FLTK anymore.
 * I don't understand why we would need Cairo. That's very odd since we only use SDL which is bitmap-based. Maybe Cairo is an FLTK dependency?

EmilieNumworks:
> * I don't understand why we would need Cairo. That's very odd since we only use SDL which is bitmap-based. Maybe Cairo is an FLTK dependency?

I don't get why but any call to `make` (like `make clean`) will output `package cairo was not found in the pkg-config` but then works correctly. So you're right, it is not needed but it triggers strange warnings...

Ecco:
I suggest we add libusb for now because it's definitely missing. And let's discuss cairo in another PR if needed 😄

Issue #1502, opened 2020-04-11
Windows size inflluence the drawing


RedGl0w:
#### Describe the bug
Try this script on simulator :
```py
from turtle import *
goto(100,100)
```

It should have some issue.

That is related to : https://github.com/numworks/epsilon/issues/1485 and https://github.com/numworks/epsilon/pull/1327#issuecomment-610838739

#### Screenshots
![image](https://user-images.githubusercontent.com/43498612/79040099-74c54000-7be6-11ea-9c11-ab2ff8443c14.png)
![image](https://user-images.githubusercontent.com/43498612/79040103-7989f400-7be6-11ea-9c52-83ab50ddc6d6.png)

#### Expected behavior
Have a straight line

#### Environment
 - Epsilon master
 - Linux and web simulator (I guess on all simulators). Not device

RedGl0w:
With https://github.com/numworks/epsilon/pull/1175, this issue doesn't appear, so maybe that's not the anti-aliasing...

RedGl0w:
The resizing of the windows influence the screen.
See : https://youtu.be/xys-sRXSqOI

artaxxx:
@RedGl0w there's no issue with the drawing of a line with turtle. The issue is maybe in a code you wrote to add a line in the kandinsky module. So let me close this issue and we'll see the problem reviewing #1511

RedGl0w:
Sorry, but it clearly appears on epsilon master with linux, like I showed on the [youtube video](youtu.be/xys-sRXSqOI), I retried and I still have the issue, and :
```
redgl0w@anarchy ~/B/epsilon (master)> git status
Sur la branche master
Votre branche est à jour avec 'origin/master'.

rien à valider, la copie de travail est propre
```
This is clearly due to the ability to change the proportion of the screen, which maybe doesn't work well on kde (and a lot of other linux DM).

artaxxx:
Ok, let me edit your first message!

PR #1598, opened 2020-06-17
[simulator] Make the calculator's screen pixel-perfect

Ecco:
Try to make the calculator's screen fit perfectly on the host's pixel grid. This is **so** much easier on the eyes 😄


RedGl0w:
This fixed #1502 :
image

Ecco:
> This fixed #1502 :

Great! Looks like the calculator screen isn't very well centered though. It was when I tried it on a Mac though, so it might be a rounding error. Maybe it'd be a good idea to shift the screen area by 1px vertically and 2px horizontally?

PR #1608, opened 2020-07-03
Added a submenu for Ion Keys in python

RedGl0w:
Added a submenu in the python toolbox for having easier chose of the key (for ion module's keydown).
Screenshots :
image
image
image
### Todo :
- [ ] Change the display of this submenu (2nd screenshot)
- [ ] Add translations
- [x] Find a better way than having an infinit loop (which can cause a lot of issue on simulator : can't close window and windows which won't be redrawn)
- [ ] Remove the valgrind file

### DIFF
--- a/build/device/dfu.py
+++ b/build/device/dfu.py
@@ -4,6 +4,8 @@
 # This work is licensed under the MIT license, see the file LICENSE for
 # details.
 
+# This script has been modified by NumWorks contributors
+
 """This module implements enough functionality to program the STM32F4xx over
 DFU, without requiring dfu-util.
 See app note AN3156 for a description of the DFU protocol.


Ecco:
That's a very interesting idea! It's probably more convenient for power user, but I'm wondering if newcomers would easily understand this. It's kinda neat though!

RedGl0w:
> That's a very interesting idea! It's probably more convenient for power user, but I'm wondering if newcomers would easily understand this. It's kinda neat though!

With a more complete interface, it would be as easy as it is currently to understand I think. Moreover, when you have understand one time how it works, it is way quicker. And this PR makes save Rom.

RedGl0w (commented):
Need to be rebased, and need to rework the interface

Issue #1547, opened 2020-05-10
Use external Flash chip for storage


boricj:
#### Problem you'd like to fix
Currently, epsilon uses a 32 KiB buffer in RAM as a storage area for Python scripts and Poincare objects (variables, sequences, functions...) with an simple ad-hoc file-system. Its structure is a sequence of (size, filename.extension, data) records with two magic `uint32_t` values delimiting the beginning and the end of the buffer. Of particular note, the workshop will directly read and write this buffer for transferring Python scripts and as such needs to understand the file-system format to work.

This file-system as it currently exists has several shortcomings:
* It has a capacity of only 32 KiB.
* Records are limited to 64 KiB in size due to the datatype used (`uint16_t`).
* Records are byte-aligned but contains POD structures that require stricter alignment, requiring tons of Emscripten-specific glue to work around.
* The file system layout makes no accommodation for Flash storage constraints (sector erasing to all ones, writing to zeroes) as the code assumes it can `memmove()` things around without issues.
* Python scripts are prefixed with a single byte denoting their auto-importation status, which makes them binary files and could potentially complicate third-party uses of these.
* The buffer uses permanently 1/8th of the MCU's total RAM that could be used for other purposes.

#### Describe the solution you'd like
The N0110 model incorporates an external 8 MiB Flash chip that is currently used for storing the firmware and the exam mode status. It is not used to its fullest potential as nearly 7 MiB of it is unused. Since it is unlikely that the firmware will expand to consume all that space anytime soon, it is proposed to move storage into it.

A new file system is required to make use of it, of which these are the requirements:
* Accommodate flash storage constraints (4 KiB block erase to ones, bits can then be set to zeroes, wear is a concern).
* Alignment to at least 4 bytes for aligned memory access.
* Greater than 64 KiB file size support.
* Contiguous storage of file data.

Having a new file system with a capacity of several megabytes instead of 32 KiB allows storing things that are currently impossible, such as images, screenshots, rich text documents, large datasets and so on. This in turn allows new usages for end-users (background images for graphs, document viewer, blitting sprites and tilesets in Python...) and also for third-party forks.

#### Describe alternatives you've considered
[...]

PR #1609, opened 2020-07-06
[mpy] Added file support

M4xi1m3:
This PR aims to add support for files in micropython. I've implemented the standards [IO functions](https://docs.python.org/3/library/io.html) in an ion.file object, and implemented the open method. This is still a subset of what a real python3 interpreter can do (bin files and text files are handled in the same class, no way to open stdout/stderr/stdin, etc.) but is still very nice. I've also added help in the toolbox for this object and for exception handling (as this throws exceptions like the real implementation). This PR has also been made on [Omega](https://github.com/Omega-Numworks/Omega/pull/366).


Ecco:
Hi @M4xi1m3 ! Thanks for this very interesting PR. It's indeed a great functionality, but it raises a major question: what kind of filesystem do we want to expose? Our current state of things uses a simplistic approach which works quite well: as you know, we use a very simple garbage-collected RAM-only stash-of-files (I don't think we can even call it a FS), and so far we're pretty happy about it. But we may want to change it at some point (for example to do stuff such as #1547).

The problem is that if we expose Python APIs to our current FS, then we would have to maintain them when we'll want to change the FS implementation.

Also, we'd like to stick as much as possible to "standard" Python, so if we were to do file I/O then we'd probably want to be able to simply [use 'open' as in the Python docs](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files). And correct me if I'm wrong, but your PR wouldn't allow exactly that, would it?

M4xi1m3:
I understand your concerns about the file system. The ion.file type is literally a wrapper around <ion/storage.h>, allowing file operations like in standard python. I did with what I had to be honest. If something like #1547 is implemented, it won't be that hard to update the ion.file type code, using the methods from the new system.

Regarding `open`, this PR implements it but without all the arguments (only name and mode). Open simply creates an ion.file type (by giving the arguments to the constructor) and returns it.

Something that would also be needed is a subset of the os module (already implemented in Omega, I can add it to this PR too) for things like remove, rename, listdir.

PR #1611, opened 2020-07-07
Removed dfu-util test

RedGl0w:
dfu-util isn't used anymore for flashing, so let's remove it (or maybe change it to test with dfu.py ?)


EmilieNumworks:
Hello @RedGl0w, I just force-pushed your change in order to relaunch the checks and rebase your commit on the current head of master.

EmilieNumworks:
We indeed don't need to test dfu-utils but, as you said, similar tests using "python3 build/device/dfu.py" instead would be great. We might also add tests to write/read the external flash when the descriptor attests that 0x90000000 is mapped, ie, when `python3 build/device/dfu.py -l` returns
```
Memory Layout
    0x8000000  4 pages of  16K bytes
    0x90000000  8 pages of   4K bytes
    0x90008000  1 pages of  32K bytes
    0x90010000 63 pages of  64K bytes
    0x90400000 64 pages of  64K bytes
```

PR #1613, opened 2020-07-09
Resolved the wrong gender of font in fr

RedGl0w:
The french submenu isn't well gendered (currently, it is written `petit`, but the submenu is called `Police python`).


artaxxx:
Thanks @RedGl0w :)

LeaNumworks:
Thanks!

Issue #929, opened 2019-03-22
[iOS] [iPhone X/Xs] Swiping up to leave the app also presses a key unexpectedly


adriweb:
#### Describe the bug
On an iPhone X/XS (Max), swiping up (there is no home button) in order to leave the app will also trigger a key press event on the key that was below the user's finger at the time of the swipe.
Basically on such devices, it would probably be good do scale everything down just a little bit so that there's extra margin at the bottom (and possibly at the top, since the notch hides the "NUMWORKS" label anyway)

#### To Reproduce
1. Go to the Calculation app
2. Swipe up to leave the Numworks app
3. `x10ˣ` for instance, will get typed as the app is going into the background

#### Expected behavior
The app goes into the background without having an unintended keypress triggered

#### Environment
iOS 12.1 on iPhone Xs

adriweb:
Apparently this also happens with some android devices with touches/gestures at the bottom of the screen.

Ecco:
Indeed it does!

rsammelson:
This is also a problem on the iPhone 11. One possible solution would be to only trigger button on a press. Sliding over multiple buttons probably shouldn’t trigger multiple presses. Then, as long as the bottom of the screen isn’t included in any hit boxes, as swipe up wouldn’t trigger any buttons.

Stephan-P:
NumWorks Android version 24.5.3 appears to have changed its use of the available screen canvas. At least on my Pixel 7a. There are now grey top and bottom margins around the NumWorks app, allowing to swipe close the app more elegantly with less chance of inadvertent key presses.

Maybe this issue could be closed now?

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/cbe7aa1d-91bd-46c4-93d6-bf605f8bd8c1" />

PR #1615, opened 2020-07-09
Visual keyboard only change state on keydown

RedGl0w:
This doesn't fix https://github.com/numworks/epsilon/issues/929, but change the behavior enough, to say that this issue can be closed. This also recreate the same behavior as the online simulator.


LeaNumworks:
Hi @RedGl0w, thanks for the PR.
I do not really get its purpose though, and it breaks some wanted behaviours.
For instance, input a computation and click for a long time on an arrow -> the cursor should move repeatedly, but with your PR it only moves once then stops.

RedGl0w:
That's an issue I noticed, and I'm going to try to fix it soon

PR #1616, opened 2020-07-11
[Ion::Keyboard::scan] Added some scancode

RedGl0w:
This is very usefull with Ion module of mpy port. Some keys aren't mapped, because they don't exist at simulator's host's keyboard (ex : sin, pi, sqrt, ...)


ion/src/simulator/shared/keyboard_sdl.cpp
```diff
@@ -36,16 +36,41 @@ class KeySDLKeyPair {
...
+  KeySDLKeyPair(Ion::Keyboard::Key::One,            SDL_SCANCODE_1),
+  KeySDLKeyPair(Ion::Keyboard::Key::Two,            SDL_SCANCODE_2),
+  KeySDLKeyPair(Ion::Keyboard::Key::Three,          SDL_SCANCODE_3),
+  KeySDLKeyPair(Ion::Keyboard::Key::Plus,           SDL_SCANCODE_KP_PLUS),
+  KeySDLKeyPair(Ion::Keyboard::Key::Minus,          SDL_SCANCODE_MINUS),
+  KeySDLKeyPair(Ion::Keyboard::Key::Minus,          SDL_SCANCODE_KP_MINUS),
+  KeySDLKeyPair(Ion::Keyboard::Key::Zero,           SDL_SCANCODE_0),
+  KeySDLKeyPair(Ion::Keyboard::Key::Dot,            SDL_SCANCODE_KP_PERIOD),
+  KeySDLKeyPair(Ion::Keyboard::Key::Dot,            SDL_SCANCODE_PERIOD),
+  KeySDLKeyPair(Ion::Keyboard::Key::EXE,            SDL_SCANCODE_RETURN),
```
RedGl0w, line 72: Maybe this could be changed to Ok,
So KP_ENTER is EXE and RETURN is OK ?

PR #1621, opened 2020-07-14
Fix: rename all the IS by SI... because even in english we say "SI" a…

coco33920:
The International System of Units is shortened "SI" in all the languages because it's french :), same as we use UTC ( which is a chimere between french&english )
It matters a lot to me, a fellow fan of the International System of Units :) 
cf: https://en.wikipedia.org/wiki/International_System_of_Units :)

TODO: well compilation :kappa:
*it's not funny when the CI is broken*

### DIFF
--- a/apps/calculation/calculation.cpp
+++ b/apps/calculation/calculation.cpp
@@ -112,7 +112,7 @@ Layout Calculation::createExactOutputLayout(bool * couldNotCreateExactLayout) {
     return PoincareHelpers::CreateLayout(exactOutput());
   } else {
     *couldNotCreateExactLayout = true;
-    return Layout(); 
+    return Layout();
   }
 }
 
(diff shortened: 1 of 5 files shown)


apps/calculation/calculation.cpp
```diff
@@ -112,7 +112,7 @@ Layout Calculation::createExactOutputLayout(bool * couldNotCreateExactLayout) {
     return PoincareHelpers::CreateLayout(exactOutput());
   } else {
     *couldNotCreateExactLayout = true;
-    return Layout();
+    return Layout();
```
boricj, line 115: Extraneous whitespace.

EmilieNumworks:
Haha, yes indeed! Thanks @coco33920 👍

PR #1633, opened 2020-07-28
Improve simulator

EmilieNumworks:
- add haptic effects on smartphones
- highlight the selected key on all simulators


ion/src/device/shared/events.cpp
```diff
@@ -0,0 +1,10 @@
+#include <ion/events.h>
+
+namespace Ion {
+namespace Events {
+
+void didPressNewKey() {
```
EmilieNumworks, line 6: Check that no jump instruction is emitted for the device

EmilieNumworks: Checked!

ion/src/simulator/android/src/cpp/haptics.cpp
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+void shutdown() {
+  if (sSDLHaptic) {
+    SDL_HapticClose(sSDLHaptic);
+  }
+}
+
+void perform() {
```
Ecco, line 28: `void rumble();`

ion/src/simulator/android/src/cpp/haptics.cpp
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+}
+
+void shutdown() {
+  if (sSDLHaptic) {
+    SDL_HapticClose(sSDLHaptic);
+  }
+}
+
```
Ecco, line 27: add a `bool isEnabled()` that is re-implemented per platform. The dummy should default to false as a failsafe (add a comment).

PR #1640, opened 2020-08-05
Execute only two times libpng-config

RedGl0w:
This is a bit cleaner, and make win one second of compile time, I think


Ecco:
Hi @RedGl0w ! I'm not sure what the point of this PR is. Would you mind giving more details? Thanks!

RedGl0w:
This is a small optimization. The flags won't be changed during the compilation, so it will be a bit more performing to get the flags at the beginning, and not for every files.

Ecco:
Actually, I'm not sure this PR would have made any difference. AFAIK, the variables (SFLAGS/LDFLAGS) are assigned with "=" instead of ":=" anyway, so the shell command would have been called every single time.

Maybe the "right" way would have been to define a specific variable (e.g. LIBPNG_FLAGS), and ":=" assign it.

But I'm not sure it's worth the pain 😄

Issue #964, opened 2019-04-26
Can't set the default language


ljoets:
Hi,
I want to set french as default language when launching the simulator.
I build epsilon.elf (on Debian Stretch) and epsilon.exe (on windows 7) files as described here : https://www.numworks.com/resources/engineering/software/sdk/, but when launching them with --language fr parameter as proposed here : https://github.com/numworks/epsilon/issues/963 The calculator still appear in english.
What's wrong ?
Thanks

RedGl0w:
The issue is that with theses [lines](https://github.com/numworks/epsilon/blob/edafa0e15555bdd024660e78e36756651bd85438/ion/src/simulator/shared/main_sdl.cpp#L24-L28), when you start the simulator with `--language fr`, the simulator will be in fr, but just after, if a default language is detected on the host, the simulator will change to the other one. Instead of `push_back`, the language should be added, but in first position.

RedGl0w:
Fixed with https://github.com/numworks/epsilon/commit/79985a019938b94a81f3de27bdc7262441b42aaf (I think)

PR #1641, opened 2020-08-05
[Ion] Reenable language as flag even with getLanguageCode()

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/964

### DIFF
--- a/ion/src/simulator/shared/main_sdl.cpp
+++ b/ion/src/simulator/shared/main_sdl.cpp
@@ -23,8 +23,8 @@ int main(int argc, char * argv[]) {
 
   char * language = IonSimulatorGetLanguageCode();
   if (language != nullptr) {
-    arguments.push_back("--language");
-    arguments.push_back(language);
+    arguments.insert(arguments.begin(), language);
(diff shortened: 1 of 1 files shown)


ion/src/simulator/shared/main_sdl.cpp
```diff
@@ -23,8 +23,8 @@ int main(int argc, char * argv[]) {
 
   char * language = IonSimulatorGetLanguageCode();
   if (language != nullptr) {
-    arguments.push_back("--language");
-    arguments.push_back(language);
+    arguments.insert(arguments.begin(), language);
+    arguments.insert(arguments.begin(), "--language");
```
EmilieNumworks, line 27: Adding "--language LANGUAGE" option only when the "--language" option is not already present in the arguments would be more robust maybe ? In the case we change how we parse the options in Epsilon.

RedGl0w: See https://github.com/numworks/epsilon/pull/1641/commits/2c42aa674b8a6f271066d11110f53d1019d17871

Issue #1643, opened 2020-08-07
White text using draw_string isn't really white/invisible.


Afyu:
#### Describe the bug
Within the Python app, white text displayed using draw_string and the 'white' colour parameter doesn't really display a white text (depending on characters). The color is very light grey. Using (255,255,255) tuple instead of 'white' colour parameter doesn't fix the issue. This issue occurs on both simulator and calculator.

#### Screenshots
![Screen](https://user-images.githubusercontent.com/69348636/89641796-c343cc00-d8b2-11ea-851e-c83ecf3a531c.png)
![white_text](https://user-images.githubusercontent.com/69348636/89641683-824bb780-d8b2-11ea-9a54-bd75980b333d.png)

#### To Reproduce
1. Go to the Python app
2. In a script, type 
from kandinsky import *
draw_string("text to display",0,0,'white')
draw_string("second text to display",0,18,(255,255,255))
3. Execute the script
4. See error. Some characters are displayed in a very light grey colour which is not white.

#### Expected behavior
The displayed text is really white, then invisible.

#### Environment
 - Epsilon version 14.4.0 on both device and simulator.
 - The platform(s) on which the problem happens: both online simulator and actual device.

sagessylu:
This is works as intended: using varied colors to display a font is an old trick that makes the font look smoothier.
This is an unexpected consequence of that behavior, but it is unavoidable in my opinion.

Ecco:
This might indeed be due to antialiasing, but it still shouldn't happen nonetheless. Thanks for the bug report @Afyu !

GabrielNumworks:
Thanks for the report. This was indeed a problem with the method used to smoothly displays fonts : because of some faulty arithmetic, the program managed to find a colour gradient between two identical colours. Pull request #1650 fixes this issue.

LeaNumworks:
Closing, as #1650 was merged

PR #1650, opened 2020-08-25
Correctly blend identical colors

GabrielNumworks:
This fixes issue #1643 : "White text using draw_string isn't really white/invisible."

### DIFF
--- a/kandinsky/src/color.cpp
+++ b/kandinsky/src/color.cpp
@@ -7,7 +7,11 @@ KDColor KDColor::blend(KDColor first, KDColor second, uint8_t alpha) {
    * dealt with. So let's make a special case for them. */
   if (alpha == 0) {
     return second;
-  } else if (alpha == 0xFF) {
+  }
+  if (alpha == 0xFF) {
+    return first;
+  }
+  if (first == second) {
     return first;
   }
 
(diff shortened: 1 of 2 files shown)


kandinsky/src/color.cpp
```diff
@@ -7,22 +7,21 @@ KDColor KDColor::blend(KDColor first, KDColor second, uint8_t alpha) {
    * dealt with. So let's make a special case for them. */
   if (alpha == 0) {
     return second;
-  } else if (alpha == 0xFF) {
+  }
+  if (alpha == 0xFF) {
+    return first;
+  }
+  if (first == second) {
```
Ecco, line 14: IIRC, those 2 cases were supposed to simply be a fast path. Is it still the case with that new 3rd one? In other words, would the new case (`first == second`) yield correct results even if this test was disabled?

kandinsky/src/color.cpp
```diff
@@ -7,22 +7,21 @@ KDColor KDColor::blend(KDColor first, KDColor second, uint8_t alpha) {
...
     return first;
   }
 
   // We want to do first*alpha + second*(1-alpha)
   // First is RRRRR GGGGGG BBBBB
   // Second is same
 
-  uint8_t oneMinusAlpha = 0xFF-alpha;
+  uint16_t oneMinusAlpha = 0x100-alpha;
```
Ecco, line 22: Good catch 😄

kandinsky/src/color.cpp
GabrielNumworks: With the change made to `oneMinusAlpha`, the case `first == second` is indeed nothing more than a shortcut. I did some benchmarking, and it seems to happen often enough for this check to yield some speed gain (albeit small when compared to the other two). I'll change the first comment to reflect that.

PR #1655, opened 2020-08-31
Move command-line argument parsing into the simulator

Ecco:
For obvious reasons we don't use command-line arguments on the actual device. Therefore this PR moves command-line argument parsing to the simulator.

* "code-lock-on-console" is now a compile-time variant, which matches the fact that we currently use this feature in a specific Epsilon.js build consisting of the Code app only.
* Introducing a new "Script::Status" class, and make it private.
* Locale lookup by ISO code (fr -> French) has been greatly simplified (since we already use those codes to designate the locale in the first place)


Ecco:
Bad design. Closing.

Issue #1637, opened 2020-08-02
Poincaré's definition for the astronomical unit has a typo


Volleo6144:
#### Describe the bug
The astronomical unit's value in metres (defined in [`poincare/include/poincare/unit.h:294`](https://github.com/numworks/epsilon/blob/master/poincare/include/poincare/unit.h#L294)) is `149587870700_m` (amusingly, `149587870700` has more than a full page of results on Google, so y'all aren't the first to make this mistake; `149597870700` has almost 300,000) instead of `149597870700_m` (the definition as per the "Astronomical unit" article on Wikipedia, which cites PDF page 3 of [this](https://www.iau.org/static/resolutions/IAU2012_English.pdf) as its source).

#### Screenshots
![_au→_m = 1.495879ᴇ11_m](https://user-images.githubusercontent.com/38385509/89112778-a419ea00-d42d-11ea-95ad-10765cc55652.png)

#### To Reproduce
1. Go to the 'Calculation' app
2. Type '_au→_m'
3. There's a typo in the answer

#### Expected behavior
That the result would be shown as 149 5**9**7 870 700 m instead of 149 5**8**7 870 700 m

#### Environment
 - Epsilon version (Settings > About > Software version): 14.4.0
 - The platform(s) on which the problem happens: I've reproduced this on both online simulator and actual device, and tracked the exact line of the error

...And yes, I am well aware that it would be difficult to have an issue of a lower priority than this.

artaxxx:
Fixed in 15.0.0

PR #1660, opened 2020-09-10
Fix Issue #1637: Incorrect Astronomical Unit's value

henryleberre:
I fixed the typo reported by issue #1637 that reported that the astronomical unit's value in meters had a typo:

> Expected behavior
> That the result would be shown as 149 597 870 700 m instead of 149 587 870 700 m

I simply modified its value in `poincare/include/poincare/unit.h:294` to be the correct one.
Thanks for creating epsilon!

### DIFF
--- a/poincare/include/poincare/unit.h
+++ b/poincare/include/poincare/unit.h
@@ -291,7 +291,7 @@ public:
         Representative("m",   nullptr,
             Representative::Prefixable::Yes,
             LongScalePrefixes),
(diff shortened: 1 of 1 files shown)


PolarToCartesian:
The macOS bug seems totally unrelated to my commit.

artaxxx:
Thanks @henryleberre , this has been fixed by https://github.com/numworks/epsilon/commit/0a92499dd797f15c6bda7a75ac1b6c869d92ae2b

Issue #1661, opened 2020-09-12
Bug that causes the calculator to freeze


ixul2:
#### Describe the bug
if you execute time.sleep(-0.01) in the python shell the calculator completely freezes. If you try it on the online simulator of the calculator, the tab crashes. The same happens if you replace "-0.01" with any negative number with at least 2 digits after the point

#### Screenshots
![crash_test](https://user-images.githubusercontent.com/44448949/93005663-62637f80-f553-11ea-915b-c8d4a2321788.png)

#### To Reproduce
1. Go to the Python app
2. Create a new script
3. write the following script:
from time import sleep
sleep(-0.01)
4. Save it
5. Execute it

#### Expected behavior
If you try to execute the same script with CPython, an exception is raised. If you execute time.sleep(-2) in the python app of the calculator it seems to produce the same result as time.sleep(0). So I guess those would be better ways of executing the line than what the calculator currently does.

#### Environment
 - Epsilon version 14.4.0.
 - The platform(s) on which the problem happens: both the physical version and the simulator

artaxxx:
Fixed

RedGl0w:
> Fixed

I guess I can close https://github.com/numworks/epsilon/pull/1662 so ?

artaxxx:
Yes!

PR #1662, opened 2020-09-13
[Python] Added "assertion" of positive time

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/1661, by adding a "cpython" asserrtion like error when time < 0

### DIFF
--- a/python/port/mod/time/modtime.cpp
+++ b/python/port/mod/time/modtime.cpp
@@ -1,17 +1,21 @@
 extern "C" {
 #include "modtime.h"
+#include <py/runtime.h>
 }
 #include <ion/timing.h>
 #include "../../helpers.h"
 #include <py/smallint.h>
-#include <py/runtime.h>
 
 mp_obj_t modtime_sleep(mp_obj_t seconds_o) {
 #if MICROPY_PY_BUILTINS_FLOAT
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_float(seconds_o));
+  mp_float_t tts = 1000 * mp_obj_get_float(seconds_o);
 #else
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_int(seconds_o));
+  mp_int_t tts = 1000 * mp_obj_get_int(seconds_o);
 #endif
+  if(tts < 0){
+    mp_raise_ValueError("sleep length must be non-negative");
+  }
+  micropython_port_interruptible_msleep(tts);
   return mp_const_none;
 }
 
--- a/python/port/mod/time/modtime.cpp
+++ b/python/port/mod/time/modtime.cpp
@@ -1,18 +1,19 @@
 extern "C" {
 #include "modtime.h"
+#include <py/smallint.h>
 #include <py/runtime.h>
 }
 #include <ion/timing.h>
 #include "../../helpers.h"
-#include <py/smallint.h>
 
 mp_obj_t modtime_sleep(mp_obj_t seconds_o) {
+  int32_t tts;
 #if MICROPY_PY_BUILTINS_FLOAT
-  mp_float_t tts = 1000 * mp_obj_get_float(seconds_o);
+  tts = 1000 * mp_obj_get_float(seconds_o);
 #else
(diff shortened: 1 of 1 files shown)


python/port/mod/time/modtime.cpp
```diff
@@ -1,17 +1,21 @@
...
 mp_obj_t modtime_sleep(mp_obj_t seconds_o) {
 #if MICROPY_PY_BUILTINS_FLOAT
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_float(seconds_o));
+  mp_float_t tts = 1000 * mp_obj_get_float(seconds_o);
 #else
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_int(seconds_o));
+  mp_int_t tts = 1000 * mp_obj_get_int(seconds_o);
 #endif
+  if(tts < 0){
+    mp_raise_ValueError("sleep length must be non-negative");
```
Ecco, line 16: I think we could simply do `mp_raise_ValueError(nullptr);` since even CPython doesn't give an actual error message:
```
>>> sleep(-2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IOError: [Errno 22] Invalid argument
```

python/port/mod/time/modtime.cpp
```diff
@@ -1,17 +1,21 @@
...
 }
 #include <ion/timing.h>
 #include "../../helpers.h"
 #include <py/smallint.h>
-#include <py/runtime.h>
 
 mp_obj_t modtime_sleep(mp_obj_t seconds_o) {
 #if MICROPY_PY_BUILTINS_FLOAT
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_float(seconds_o));
+  mp_float_t tts = 1000 * mp_obj_get_float(seconds_o);
```
Ecco, line 11: I think `tts` should be declared as an `int32_t` since it is what `micropython_port_interruptible_msleep`eventually expects. Also `tts` should probably be defined outside of the #if block.

python/port/mod/time/modtime.cpp
```diff
@@ -1,17 +1,21 @@
...
 #else
-  micropython_port_interruptible_msleep(1000 * mp_obj_get_int(seconds_o));
+  mp_int_t tts = 1000 * mp_obj_get_int(seconds_o);
 #endif
+  if(tts < 0){
```
Ecco, line 15: Our coding style would usually be `  if (tts < 0) {`

python/port/mod/time/modtime.cpp
```diff
@@ -1,17 +1,21 @@
 extern "C" {
 #include "modtime.h"
+#include <py/runtime.h>
 }
 #include <ion/timing.h>
 #include "../../helpers.h"
 #include <py/smallint.h>
-#include <py/runtime.h>
```
Ecco, line 7: We may want to either move all the `py/*.h` to the `extern "C"` block, or just leave them as is. I'd vode for the latter 😄

Ecco (commented):
That's a neat patch, thanks! I made a few comments which aren't critical but which would make for a better patch IMHO.

python/port/mod/time/modtime.cpp
RedGl0w: Cpython raise this exact error when time < 0 indeed (https://github.com/python/cpython/blob/95a8a0e01d1f4f49ad6b4d4f2ae3bbf93c462013/Modules/timemodule.c#L337)

python/port/mod/time/modtime.cpp
RedGl0w: I need the raiseValueError, which is C linked, so I'll move every `py/*.h` in the `extern "C"`

python/port/mod/time/modtime.cpp
RedGl0w: Done with [`4de2bc3` (#1662)](https://github.com/numworks/epsilon/pull/1662/commits/4de2bc3367126e5403be94d2d7c3b10b8afdb749)

RedGl0w:
Mhh
I can't understand why is the macOS CI failing

Issue #1672, opened 2020-10-06
The sequences app can hang when computing a sum of terms


Ecco:
#### Describe the bug
The sequences app can hang when computing a sum of terms

#### Screenshots
![screenshot (12)](https://user-images.githubusercontent.com/55989/95216833-84ea6080-07c0-11eb-8681-f764541dfba8.png)

#### To Reproduce
Steps to reproduce this behavior

1. Open the menu and start the 'Sequences' application.
2. Add an 'Explicit expression' sequence.
3. Make the formula 'two to the nth power'.
4. Plot the graph of that function.
5. While looking at the graph of the sequence, press 'OK' and select 'Sum Of Terms'.  
6. Select zero as the first term and click 'OK'.
7. Keep clicking the right arrow to select the last term.
8. The calculator will lock up upon reaching 17.

#### Environment
 - Epsilon version 14.4.0
 - On the actual device and web simulator

EmilieNumworks:
Fixed by #1673

PR #1673, opened 2020-10-06
[apps/shared] CurveView: avoid overflowing KDCoordinate

EmilieNumworks:
This fixes the following bug: add the sequence, u(n)=2^n. On the graph,
go to "sum of terms" and try to sum terms from 0 to 18 by moving the
cursor. It hangs.


EmilieNumworks:
Fixes #1672

PR #1696, opened 2020-10-28
[Apps/main] Removed strcmp (see #1695)

### LINKED ISSUE #1695: Epsilon doesn't handle arguments because of strcmp
#### Describe the bug
My computer uses an implementation of `strcmp()` which will return -1 if the string is lower, 0 if it is egal, or 1 greater (and it corresponds to the [c norm](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf)).
image
The issue of that is [epsilon main](https://github.com/numworks/epsilon/blob/e0dcdcea286f7f31b9fbee4d43456a2292ac5bc8/apps/main.cpp#L55-L56) wants a function which will return the difference of chars (like it is implemented [here](https://github.com/numworks/epsilon/blob/e0dcdcea286f7f31b9fbee4d43456a2292ac5bc8/liba/src/strcmp.c#L3-L9)).
The execution arguments doesn't work at all for me, because of that requirement.

#### Screenshots
By adding before [this line](https://github.com/numworks/epsilon/blob/e0dcdcea286f7f31b9fbee4d43456a2292ac5bc8/apps/main.cpp#L56) that line :
```cpp
std::cout << argv[i] << " " << appNames[j] << " " << cmp << std::endl;
```
here is what is outputed :
image

#### Expected behavior
I would expect args to work.

#### Environment
 - Epsilon master
 - The platform(s) on which the problem happens: arch linux with gcc 10.2.0

RedGl0w:
Fixed #1695

### DIFF
--- a/apps/main.cpp
+++ b/apps/main.cpp
@@ -52,8 +52,10 @@ void ion_main(int argc, const char * const argv[]) {
     const char * appNames[] = {"home", EPSILON_APPS_NAMES};
     for (int j = 0; j < AppsContainer::sharedAppsContainer()->numberOfApps(); j++) {
       App::Snapshot * snapshot = AppsContainer::sharedAppsContainer()->appSnapshotAtIndex(j);
-      int cmp = strcmp(argv[i]+2, appNames[j]);
-      if (cmp == '-') {
+      const char * s1 = argv[i]+2;
+      const char * s2 = appNames[j];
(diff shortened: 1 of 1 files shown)


RedGl0w:
Adapting Standart to a specific use is maybe not the best solution. If `strcmp` doesn't exist to do that on specific platform, another solution using Standarts shared on all platform should be used

PR #1698, opened 2020-10-30
Small changes to German translation

vardai:
There were several expressions that needed to be corrected.


vardai:
Found several other text passages that I'm also going to translate.

EmilieNumworks:
Thanks for your help!

PR #1701, opened 2020-10-31
[workflows] Change actions/github to actions/github-script

RedGl0w:
The actions/github is deprecated (see https://github.com/actions/github) and it doesn't work anymore (https://github.com/numworks/epsilon/pull/1699/checks?check_run_id=1335601562/checks?check_run_id=1335601562)

### DIFF
--- a/.github/workflows/metrics-workflow.yml
+++ b/.github/workflows/metrics-workflow.yml
@@ -29,8 +29,7 @@ jobs:
       - name: Prepare comment auth
         run: echo "::set-env name=GITHUB_TOKEN::$(echo ZGExNWM1YzNlMjVkMWU5ZGFmOWQyY2UxMmRhYjJiN2ZhMWM4ODVhMA== | base64 --decode)"
       - name: Add comment
-        uses: actions/github@v1.0.0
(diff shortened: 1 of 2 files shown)


Ecco:
Well, apparently we should use [this](https://github.com/actions/github-script). They have an example on how to comment on an issue.

RedGl0w:
Yep I've experimented with this (that's the reason of the message of numworks bot upper), but I had an issue with escaping the chars of the output of the compare (mainly the LF). I'll try more with that

RedGl0w:
> Indeed, thanks for spotting this deprecation. I'd rather use GitHub's own tools, since those will probably be more popular and (hopefully?) more stable. Apparently they have a [new action](https://github.com/actions/github-script) for this, and they have an example on how to comment on an issue.

Changes in `binary_size.py` were needed : that was the main reason why I didn't want to use github-script. These changes are made by https://github.com/numworks/epsilon/pull/1701/commits/277bad1792abde4167bbd73badad695bda4dcfa5

RedGl0w:
Closed by #1716

PR #1707, opened 2020-11-16
module removal

Cyber-Thibaut:
Hello,
I have installed modules (such as the Nitendo emulator) that I no longer need today and would like to remove them. Unfortunately I can't find how to do it. I did the restoration of my device but nothing goes away. Can you tell me how to do it?
Thank you in advance for your help.


Cyber-Thibaut:
you just have to make a vacuum installation

PR #1708, opened 2020-11-17
[workflow] Use msys2/setup-msys2@v2

RedGl0w:
This fixes the Windows CI


EmilieNumworks:
This has been fixed in the version-15 (in a quite similar way) 👍

PR #1714, opened 2020-11-20
CI fix for Windows and MacOS

HugoNumworks:
This PR fixes the CI for Windows and MacOS


Ecco:
Isn't msys2 now part of the standard GitHub image? Do we really need an msys2 action?

EmilieNumworks:
@HugoNumworks said:
> According to https://github.com/msys2/setup-msys2 and my tests, although it is available by default, MSYS2/MINGW are neither added to the PATH nor available as a custom shell option.

Issue #1629, opened 2020-07-18
[ci] Leading zeroes get removed from NumWorksBot comments


M4xi1m3:
#### Describe the bug
Leading zeroes in 3 digits groups get removed. This can be seen in #1615 in .data, #1563 in .bss and many others

#### Screenshots
![image](https://user-images.githubusercontent.com/57234558/87859809-3ef7cc00-c938-11ea-95b2-efe2dc442e96.png)

PR #1716, opened 2020-11-23
[github/workflows] Update Metrics to remove NumWorksBot

HugoNumworks:
Change-Id: Icfd19cbad5472f6bf81a2a1accb88e79ab7a115b

Triggering on pull_request_target event, the base version of Metrics will always run on future PR, preventing misuse of this workflow.

Additionally, comment is generated using github-script API. We don't need NumWorksBot anymore.
image

### DIFF
--- a/.github/workflows/metrics-workflow.yml
+++ b/.github/workflows/metrics-workflow.yml
@@ -1,5 +1,5 @@
 name: Metrics
-on: [pull_request]
+on: [pull_request_target]
 
 jobs:
   binary-size:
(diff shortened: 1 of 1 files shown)


quentinguidee:
Stop me if I'm wrong, I haven't tested with your Workflow, but I believe it won't work. The `pull_request_target` is only triggered by PRs that are performed from a branch of the upstream repo to a branch of the upstream repo (i.e. in 0% of cases). The github secrets are indeed not accessible when a fork makes a PR here, for security reasons.

The technique to solve this problem is to keep NumWorksBot, keep `pull_request` instead of `pull_request_target` and replace the `${{ secrets.GITHUB_TOKEN }}` with the token of NumWorksBot.

RedGl0w:
By the way, it closed https://github.com/numworks/epsilon/issues/1629

Issue #1742, opened 2021-01-02
AttributeError: 'array.array' object has no attribute 'fromstring' during flash


raccoondude:
Currently when flashing, an error is thrown, this happens during compiling

The error is

```Makefile:29: HOME_DISPLAY_EXTERNALS is set but external isn't included, ignoring flag.
DFU     epsilon_flash
INFO    About to flash your device. Please plug your device to your computer
        using an USB cable and press the RESET button the back of your device.
Mass erase...
Traceback (most recent call last):
  File "/home/owo/Omega/build/device/usb/core.py", line 1023, in ctrl_transfer
    buff = util.create_buffer(data_or_wLength)
  File "/home/owo/Omega/build/device/usb/util.py", line 149, in create_buffer
    return array.array('B', _dummy_s * length)
TypeError: can't multiply sequence by non-int of type 'str'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/owo/Omega/build/device/usb/_interop.py", line 79, in as_array
    return array.array('B', data)
TypeError: cannot use a str to initialize an array with typecode 'B'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/owo/Omega/build/device/dfu.py", line 598, in <module>
    main()
  File "/home/owo/Omega/build/device/dfu.py", line 582, in main
    mass_erase()
  File "/home/owo/Omega/build/device/dfu.py", line 163, in mass_erase
    __dev.ctrl_transfer(0x21, __DFU_DNLOAD, 0, __DFU_INTERFACE,
  File "/home/owo/Omega/build/device/usb/core.py", line 1025, in ctrl_transfer
    buff = _interop.as_array(data_or_wLength)
  File "/home/owo/Omega/build/device/usb/_interop.py", line 84, in as_array
    a.fromstring(data) # deprecated since 3.2
AttributeError: 'array.array' object has no attribute 'fromstring'
make: *** [build/targets.device.n0100.mak:7: epsilon_flash] Error ```

Im using Python 3.9.1

Im very confused as to why this is happening :/

**Comment by cybfr** [comment]
Had the same error :+1: 
```
xx@xxx:~/ws/numworks/epsilon$ make MODEL=n0100 epsilon_flash
DFU     epsilon_flash
INFO    About to flash your device. Please plug your device to your computer
        using an USB cable and press the RESET button the back of your device.
Mass erase...
Traceback (most recent call last):
[...]

PR #1743, opened 2021-01-02
[pyUSB] Update and fix on python 3.9

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/1742

### DIFF
--- a/build/device/usb/__init__.py
+++ b/build/device/usb/__init__.py
@@ -1,17 +1,32 @@
-# Copyright (C) 2009-2017 Wander Lairson Costa
-# Copyright (C) 2017-2018 Robert Wlodarczyk
+# Copyright 2009-2017 Wander Lairson Costa
+# Copyright 2009-2021 PyUSB contributors
 #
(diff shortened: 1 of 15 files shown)


build/device/usb/__init__.py
```diff
@@ -1,17 +1,32 @@
-# Copyright (C) 2009-2017 Wander Lairson Costa
```
EmilieNumworks, line 1: We would rather update pyUSB with one of the released version: https://github.com/pyusb/pyusb/tags

EmilieNumworks (commented):
Thanks for fixing this! Could you rebase your PR to reduce it to the two last commits? I can do it if you're ok with me pushing on your fork.

RedGl0w:
> Thanks for fixing this! Could you rebase your PR to reduce it to the two last commits? I can do it if you're ok with me pushing on your fork.

You can do it :p
I don't have any computer yet, so do it if you want to

build/device/usb/__init__.py
RedGl0w: Indeed, good idea

Issue #1760, opened 2021-01-27
Defining the variable n with a matrix causes a bug in the display of Un in the Toolbox


A2drien:
I think showing this is enough to understand the problem. (Look below the u.)
![image](https://user-images.githubusercontent.com/52292872/106040981-82977380-60db-11eb-970f-10d0b5e6533f.png)

#### To Reproduce
1. Define the variable n like this:
![image](https://user-images.githubusercontent.com/52292872/106043077-30a41d00-60de-11eb-93ad-2a6a57fe75b8.png)

2. Go to **Sequences** and set Un = n
3. Return to **Calculations**, open the toolbox and go to sequences: contemplate the aberration.

#### Expected behavior
I'm just supposed to see _Un_ on the left.
 
- Epsilon version 15.3.0.
- N100, Simulator

Ecco:
Good catch, thanks!

RedGl0w:
Indeed, that's not specific to matrix. The issue is that in the sequence formula the n is replaced by the symbol stored in n

PR #1765, opened 2021-01-28
[Variable box] Do not replace symbol in sequence name

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/1760


GabrielNumworks:
Thanks for the PR ! This indeed fixes the issue; although I think we could avoid it altogether by using `Sequence::definitionName`, which directly creates a Layout, instead of parsing the name.

There is also another method `Sequence::nameLayout` ; the difference with `definitionName` is that the latter will include the `+1` or `+2` when applied to recursive sequences. It is not apparent in the header though. If you decide to use `definitionName`, it could be wise to add a comment about this difference. For now, the rather abstract section titles "Definition" and "Sequence properties" are all the documentation there is...

Issue #1778, opened 2021-02-08
Building simulator on W10 fails


ljoets:
Hello,

I scrupulously follow the doc : https://www.numworks.com/resources/engineering/software/build/ to build a Windows version of the simulator. (I've been successful on GNU/Linux Debian Buster). But it fails :-( 

```
Maintenance@DESKTOP-3DP1JBR MSYS ~/epsilon
$ make PLATFORM=simulator clean
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'freetype2' found
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'freetype2' found
build/targets.simulator.windows.mak:2: *** mixed implicit and normal rules: deprecated syntax
CLEAN
```
```
Maintenance@DESKTOP-3DP1JBR MSYS ~/epsilon
$ make PLATFORM=simulator epsilon_run
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'freetype2' found
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'freetype2' found
build/targets.simulator.windows.mak:2: *** mixed implicit and normal rules: deprecated syntax
make: *** No rule to make target 'epsilon_run'.  Stop.
```
What am I missing ?
Thanks

RedGl0w:
> What am I missing ?
Indeed nothing.
I opened #1779 to fix the misstype

Ecco:
Indeed, this was because of a stupid typo. Thanks @RedGl0w for fixing this! @ljoets you should be good now!

ljoets:
Hello,
I tried again today, and still the same issue.
```
admin@hppd400sff-59 MSYS ~/epsilon
# make PLATFORM=simulator clean
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'freetype2' found
Package freetype2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `freetype2.pc'
to the PKG_CONFIG_PATH environment variable
[...]

PR #1779, opened 2021-02-08
[makefiles] Fix epsilon_run on windows

RedGl0w:
This fixes https://github.com/numworks/epsilon/issues/1778


Ecco:
Wow, that was a very stupid typo! Thanks for the fix!

PR #1781, opened 2021-02-12
[Poincare] Added more talktative tests

RedGl0w:
Added on calculation_store tests some usefull infos :
```
TEST FAILURE WHILE TESTING:
  1/2
  has as DisplayOutput
  ExactAndApproximate
  instead of
  ApproximateOnly
  ```
  ```
  TEST FAILURE WHILE TESTING:
  1/0
  has as exact output
  undef
  instead of
  a
  ```
  
#### context
It is very usefull to correct the unit tests of omega
And for new contributor it will be way easier to understand why the CI fails

### DIFF
--- a/apps/calculation/test/calculation_store.cpp
+++ b/apps/calculation/test/calculation_store.cpp
@@ -83,18 +83,42 @@ QUIZ_CASE(calculation_ans) {
 void assertCalculationIs(const char * input, DisplayOutput display, EqualSign sign, const char * exactOutput, const char * displayedApproximateOutput, const char * storedApproximateOutput, Context * context, CalculationStore * store) {
   store->push(input, context, dummyHeight);
   Shared::ExpiringPointer<::Calculation::Calculation> lastCalculation = store->calculationAtIndex(0);
-  quiz_assert(lastCalculation->displayOutput(context) == display);
+
+  static_assert(static_cast<int>(DisplayOutput::Unknown) == 0 &&
+                static_cast<int>(DisplayOutput::ExactOnly) == 1 &&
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/calculation/test/calculation_store.cpp [hunk 1]
```diff
@@ -83,18 +83,42 @@ QUIZ_CASE(calculation_ans) {
...
-    quiz_assert_print_if_failure(strcmp(lastCalculation->exactOutputText(), exactOutput) == 0, input);
+    assert_compare_string(input, lastCalculation->exactOutputText(), exactOutput, "exact output");
   }
   if (displayedApproximateOutput) {
-    quiz_assert_print_if_failure(strcmp(lastCalculation->approximateOutputText(NumberOfSignificantDigits::UserDefined), displayedApproximateOutput) == 0, input);
+    assert_compare_string(input, lastCalculation->approximateOutputText(NumberOfSignificantDigits::UserDefined), displayedApproximateOutput, "displayed approximate output");
   }
   if (storedApproximateOutput) {
-    quiz_assert_print_if_failure(strcmp(lastCalculation->approximateOutputText(NumberOfSignificantDigits::Maximal), storedApproximateOutput) == 0, input);
+    assert_compare_string(input, lastCalculation->approximateOutputText(NumberOfSignificantDigits::Maximal), storedApproximateOutput, "storedf Approximate Output");
```
RedGl0w, line 121:
```suggestion
    assert_compare_string(input, lastCalculation->approximateOutputText(NumberOfSignificantDigits::Maximal), storedApproximateOutput, "stored Approximate Output");
```

PR #1784, opened 2021-02-16
Fix #1777 and #1276

### LINKED ISSUE #1276: Python app crashes in linux simulator.
#### Describe the bug
Crash in linux simulator when using python app in `Code::PythonTextArea::ContentView::drawLine` (apps/code/python_text_area.cpp:117).

#### Screenshots
https://pastebin.com/6Wqr5t18

#### To Reproduce
1. Go to the python app
2. Open a script
3. Observe a SIGILL/SEGV (depends of the situation)

#### Environment
 - Epsilon version: master (de603bef)
 - The platform(s) on which the problem happens: linux simulator. Wasn't able to test other platforms.

### LINKED ISSUE #1777: apps_layout.h not found during compilation
#### Describe the bug
Build fails when I compile Epsilon with binpack: 
`ACCEPT_OFFICIAL_TOS=1 make binpack -j8`

Output:
```
CXX     apps/home/app.o
CXX     apps/home/app_cell.o
CXX     apps/home/controller.o
apps/home/controller.cpp:3:10: fatal error: apps/home/apps_layout.h: No such file or directory
    3 | #include <apps/home/apps_layout.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [output/release/device/n0110/apps/home/controller.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [binpack] Error 2
```

I should note that I do not get this error during normal compilation (without binpack). This makes me think that the recent change to a dynamically generated apps_layout.h by apps_layout.py might not have been implemented correctly in the binpack make target, but I do not know enough about Makefiles and the Epsilon toolchain to solve this problem myself. Looking at the commits concerning this change, I think @GabrielNumworks might be able to help with this.

#### Environment
 - I'm using a Mac based on Apple's M1 ARM chip, but I do not think that is relevant to this issue.

M4xi1m3:
This fixes #1777 by adding apps_layout.h as a dependency and fixes #1276 by enabling setjmp in debug mode. This is less optimized than what is in place using inline assembly, but is enabled only on debug and makes it not crash.

### DIFF
--- a/python/port/mpconfigport.h
+++ b/python/port/mpconfigport.h
@@ -144,3 +144,10 @@ extern const struct _mp_obj_module_t modturtle_module;
     { MP_ROM_QSTR(MP_QSTR_time), MP_ROM_PTR(&modtime_module) }, \
     { MP_ROM_QSTR(MP_QSTR_turtle), MP_ROM_PTR(&modturtle_module) }, \
 
+// Enable setjmp in debug mode. This is to avoid some optimizations done
+// specifically for x86_64 using inline assembly, which makes the debug binary
+// crash with an illegal instruction
+#ifndef NDEBUG
+  #define MICROPY_NLR_SETJMP 1
+#endif
+
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -88,6 +88,7 @@ $(BUILD_DIR)/apps/i18n.h: $(BUILD_DIR)/apps/i18n.cpp
 $(eval $(call depends_on_image,apps/title_bar_view.cpp,apps/exam_icon.png))
 
 $(call object_for,$(apps_src) $(tests_src)): $(BUILD_DIR)/apps/i18n.h
+$(call object_for,$(apps_src) $(tests_src)): $(BUILD_DIR)/apps/home/apps_layout.h
 $(call object_for,$(apps_src) $(tests_src)): $(BUILD_DIR)/python/port/genhdr/qstrdefs.generated.h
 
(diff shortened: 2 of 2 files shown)


python/port/mpconfigport.h
```diff
@@ -144,3 +144,10 @@ extern const struct _mp_obj_module_t modturtle_module;
     { MP_ROM_QSTR(MP_QSTR_time), MP_ROM_PTR(&modtime_module) }, \
     { MP_ROM_QSTR(MP_QSTR_turtle), MP_ROM_PTR(&modturtle_module) }, \
 
+// Enable setjmp in debug mode. This is to avoid some optimizations done
+// specifically for x86_64 using inline assembly, which makes the debug binary
+// crash with an illegal instruction
+#ifndef NDEBUG
+  #define MICROPY_NLR_SETJMP 1
+#endif
```
EmilieNumworks, line 152: I'm not sure this is the right way to fix it - but I'm not sure I get where the crash comes from though. The illegal emitted instructions you're talking about are those in `python/src/py/nlrx64.c`? We might want to fix them and PR micropython with the fix?

M4xi1m3: Yes, I did a lot of debugging and the problem is that these instructions do a context switch, and for some reason the debug binary doesn't manage to handle that properly. It might be an upstream issue.

LeaNumworks:
@M4xi1m3, I took the liberty of cherry-piking your second commit and put it in a separate PR.

PR #1786, opened 2021-02-21
New icons for the Windows and macOS simulator

ArtichOwO:
New icons for the Windows simulator (it didn't have one before) and the macOS simulator to be in "Big Sur style" (the Windows one is still a square tho)


Ecco:
Thanks for the PR @ArtichOwO ! The icons shall all be generated from `ion/src/simulator/assets/logo.svg` using ImageMagick. See how we do it for iOS and Android.

ArtichOwO:
Thanks for your answer!
I did made a different image for each simulator, because Apple does use a different "style" for its devices, but if you want to use a svg, there's no problem, but there will still be two different files..
The Windows simulator didn't had an icon, but you're right, it should be generated from the svg; but where should I add the .ico generation command in the Makefiles ?  
Have a great day!

Issue #1678, opened 2020-10-12
App does not capture KP_Enter.


fuzebox1:
#### Describe the bug
App does not capture KP_Enter.

#### To Reproduce
The app captures the keyboards Return, but not keypads return/enter.
xev outputs keyboard enter as Return, and keypad enter as KP_Enter.

#### Expected behavior
I would expect KP_Enter to be bound in the same way Return is.

#### Environment
Episilon (and omega for that matter) -git master
Arch Linux Xorg.

RedGl0w:
Could be added to #1616

PR #1788, opened 2021-02-22
[bug] [simulator/shared] Fix numeric enter key

AntoninLoubiere:
On simulator, (not the web simulator) (tested only on Arch Linux), the enter key of the numpad don't fire the EXE key.

Fix that by pairing `SDL_SCANCODE_KP_ENTER` and `Ion::Keyboard::Key::EXE`.


RedGl0w:
Fix https://github.com/numworks/epsilon/issues/1678 and is in #1616 too

Issue #1825, opened 2021-04-11
Graphical issue with exam mode popup


RedGl0w:
#### To Reproduce
1. Enter exam mode
2. Connect USB (the exam popup opens)
3. Disconnect USB
4. Recconect USB (you should enter DFU mode)
5. Press back
6. Disconnect and reconnect USB

#### Expected behavior
I would expect the text of the popup to be drawn (or when I disconnect USB I would expect the exam popup to close ?)

#### Environment
 - Epsilon 15
 - The platform(s) on which the problem happens:  n0110 (and probably n0100)

Acssiohm:
This bug also happens, when you :
- Click on the Exam mode button ( the popup shows up )
- Connect USB ( should enter DFU mode )
- Disconnect USB
- Go again to click on the exam mode button, the popup shows up again, but incorrectly

artaxxx:
This behavior has been changed in version 16

PR #1826, opened 2021-04-12
[Exam_pop_up] Exit on USBPlug (or USBUnplug)

RedGl0w:
Fixes https://github.com/numworks/epsilon/issues/1825

### DIFF
--- a/apps/exam_pop_up_controller.cpp
+++ b/apps/exam_pop_up_controller.cpp
@@ -29,6 +29,15 @@ ExamPopUpController::ExamPopUpController(ExamPopUpControllerDelegate * delegate)
 {
 }
 
+bool ExamPopUpController::handleEvent(Ion::Events::Event event) {
+  // USBPlug is also sent with unplug
+  if (event == Ion::Events::USBPlug) {
+    Container::activeApp()->dismissModalViewController();
+    return true;
+  }
+  return PopUpController::handleEvent(event);
+}
+
 void ExamPopUpController::setTargetExamMode(GlobalPreferences::ExamMode mode) {
   m_targetExamMode = mode;
   for (int i = 0; i < k_numberOfLines; i++) {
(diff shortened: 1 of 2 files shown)


GabrielNumworks:
Nice catch ! Going with the solution to close the pop-up when the cable is unplugged is a good call. As it stands, there is just one issue.
The pop-ups to deactivate and reactivate the exam mode are actually the same, so bringing up the "Reactivate exam mode" pop-up and then plugging the device would close the pop-up without prompting the user for deactivation since you dismiss the modal view controller on plugging and unplugging indifferently.

You could fix that by adding `&& !Ion::Usb::isPlugged()` as a condition in the handleEvent, but then instead of simply closing the reactivation pop-up, plugging the device would cause the deactivation pop-up to appear, and then shortly after the device would switch to DFU as `ExamPopUpController::viewDidDisappear` is called.
So fixing this corner case is not that trivial and requires a bit more work...

RedGl0w:
> The pop-ups to deactivate and reactivate the exam mode are actually the same, so bringing up the "Reactivate exam mode" pop-up and then plugging the device would close the pop-up without prompting the user for deactivation since you dismiss the modal view controller on plugging and unplugging indifferently.

I just realized it after having opened the PR, but I didn't fix it...
> You could fix that by adding `&& !Ion::Usb::isPlugged()` as a condition in the handleEvent, but then instead of simply closing the reactivation pop-up, plugging the device would cause the deactivation pop-up to appear, and then shortly after the device would switch to DFU as `ExamPopUpController::viewDidDisappear` is called.

I'm not sure to understand well what you're talking about. The fix I was thinking about is `&& m_targetExamMode == GlobalPreferences::ExamMode::Off` (but there's still a strange behavior when you connect USB during activate / reactivate, which makes the current popup disappear, if it is the reactivate popup previously, makes appear the desactivate exam mode, and then enter in DFU mode).
I'm also trying to fix the display issue, which can still happens.

Issue #1831, opened 2021-04-22
Error when computing a finite interval with a value assigned to x


Ecco:
#### Describe the bug
The output of `int(x+1/x,x,1,2)` changes if `x` is given a value.

#### Screenshots
![screenshot (49)](https://user-images.githubusercontent.com/55989/115785258-3764c800-a38d-11eb-925d-f6a568492758.png)
![screenshot (50)](https://user-images.githubusercontent.com/55989/115785260-392e8b80-a38d-11eb-84d9-9821c11dc009.png)

#### To Reproduce
See https://workshop.numworks.com/simulators/y573zk

#### Environment
 - Epsilon version 15
 - Every platform

Ecco:
It seems like `x+1/x` is turned into `(x^2+1)/x` (that's ok), and then the `x` in the denominator is turned into a `1`, but not the one on the numerator. This yields `int(x^2+1,x,1,2)` which is indeed close to `3.3333`. If you put 2 into `x` then you get `1.66667` which would be compatible with this hypothesis.

Ecco:
Instrumenting `Evaluation<T> IntegralNode::templatedApproximate(ApproximationContext approximationContext) const` shows that this is indeed what's going on.

artaxxx:
@HugoNumworks , it's fixed right?

HugoNumworks:
@artaxxx yes, it is fixed for V-16 !

PR #1832, opened 2021-04-23
[poincare/addition] Fix issue with symbolic computation

HugoNumworks:
This PR fixes https://github.com/numworks/epsilon/issues/1831

When factorizing on a common denominator, the temporary expressions don't have parents, and don't know about `x`'s definition from the integral.

Therefore, symbols should not be replaced with their definition here.

I also added a unit tests.

### DIFF
--- a/poincare/src/addition.cpp
+++ b/poincare/src/addition.cpp
@@ -436,19 +436,17 @@ Expression Addition::factorizeOnCommonDenominator(ExpressionNode::ReductionConte
    * parent here), which would be replaced with undef.
    * Example: int((ℯ^(-x))-x^(0.5), x, 0, 3), when creating the common
    * denominator for the integrand. */
-  ExpressionNode::ReductionContext contextWithSymbolicComputation = ExpressionNode::ReductionContext(
-      reductionContext.context(),
-      reductionContext.complexFormat(),
-      reductionContext.angleUnit(),
-      reductionContext.unitFormat(),
-      reductionContext.target(),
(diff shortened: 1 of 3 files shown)


poincare/src/addition.cpp
```diff
@@ -429,11 +429,11 @@ Expression Addition::factorizeOnCommonDenominator(ExpressionNode::ReductionConte
...
-   * disconnected from the parametered expression, which would be replaced with
-   * undef.
+  /* To simplify the numerator and the denominator, we temporarily disable
+   * symbolic computation: all unwanted symbols should have already disappeared
+   * by now, and if we checked again for symbols we might find "parameter"
+   * symbols disconnected from the parametered expression (which is no longer a
+   * parent here), which would be replaced with undef.
    * Example: int((ℯ^(-x))-x^(0.5), x, 0, 3), when creating the common
    * denominator for the integrand. */
   ExpressionNode::ReductionContext contextWithSymbolicComputation = ExpressionNode::ReductionContext(
```
RedGl0w, line 439: The variable should probably be renamed

LeaNumworks, line 439: You can also use a setter (to define?).
```cpp
ExpressionNode::ReductionContext contextWithoutSymbolicComputation = reductionContext;
contextWithoutSymbolicComputation.setSymbolicComputation(ExpressionNode::SymbolicComputation::DoNotReplaceAnySymbol);
```

LeaNumworks (approved):

PR #1800, opened 2021-03-11
[python] Add a draw_line function in kandinsky

Lauryy06:
Like #1511, but the function is in the toolbox, and there isn't any conflict.

PR #1841, opened 2021-05-10
[Python/kandinsky] Added some useful functions in kandinsky module

Lauryy06:
This is a rework of #1800 which added a `draw_line()` function in the python kandinsky module, but with 3 new functions added:
 - `draw_circle()`
 - `fill_circle()`
 - `fill_polygon()`

Here is a simple example, this code 
image
do this
image


cent20net:
We need many useful functions in kandinsky ! Love it !

Acssiohm (changes requested):

kandinsky/src/context_polygon.cpp
```diff
@@ -0,0 +1,68 @@
...
+            if (switchesX[i]>=right) 
+                break;
+
+            if (switchesX[i+1]>left) {
+                if (switchesX[i]<left) 
+                    switchesX[i]=left;
+                if (switchesX[i+1]>right) 
+                    switchesX[i+1]=right;
+                for (int x=switchesX[i]; x<=switchesX[i+1]; x++) 
+                    setPixel(KDPoint(x,y), color);
```
Acssiohm, line 63: Same as the circle, you can use fillRect to draw a horizontal line :
fillRect( KDRect( switchesX[ i ] , y, switchesX[ i+1 ], 1 ) ;
and you don't have to verify if switchesX[i] < left and switchesX[i+1] > right , because fillRect already normalize the result

RedGl0w (changes requested):
You should probably respect the code style of epsilon in `kandinsky/src/context_polygon.cpp` (brackets not on a new line, with a space before, 2 spaces for indentation, no if without brackets, and a blank line at the end).

Lauryy06:
I added a commit following your advices, thanks !

kandinsky/src/context_polygon.cpp
```diff
@@ -4,65 +4,66 @@
...
-                    setPixel(KDPoint(x,y), color); 
-            }
-        }
+      if (switchesX[i+1]>left) {
+          fillRect( KDRect( switchesX[ i ] , y, switchesX[ i+1 ] - switchesX[ i ], 1 ), color ) ;
+      }
     }
+  }
 
 }
```
RedGl0w, line 69:
```suggestion
} 

```

Lauryy06: Would you advise me to add this to my last commit and force-push, revert the last commit and redo it with this addition, or make a new commit with just this?

RedGl0w: Tou should probably commit that, and then squash it with the previous commit

Lauryy06:
That's it, I have squashed all the commits into a single one.

PR #1856, opened 2021-06-17
Enhance Unit Conversion, rename _gal to _galUS, and add _galUK

edwardtianye:
The default gallon is of US standard. 
So I added UK standard as _galUK, and renamed _gal to _galUS. 
And revised the toolbox tips accordingly.


artaxxx:
Thanks for your PR @edwardtianye ! Let me suggest some changes:
* `_gal` should remain `_gal` and the legend "Gallon"
* let's add `_galUK` with the legend "British gallon"
* results or additional results should be given in `_gal` (not in `_galUK`), except if it's an explicit conversion: `1_L ⇾ _galUK` for instance

edwardtianye:
Thanks for your prompt response, per to you suggestion, have the changes done.

edwardtianye:
Sorry, should make a PR on version-15 branch?

GabrielNumworks (changes requested):
Thanks for the PR !
I tested it and found that `_galUK` appears in results instead of `_gal`. To fix that, you'd need to change the number of authorized volume representative in simplification. This is done in poincare/src/unit.cpp, in the definition of `VolumeRepresentative::standardRepresentative` : the fourth argument of the call to `DefaultFindBestRepresentative` needs to be changed from `numberOfRepresentatives() - 1` to `numberOfRepresentatives() - 2`.
Appart from that and the small style comments, this is good for me !

PR #1862, opened 2021-08-04
Base the 'msleep' function on MillisElapsed

Acssiohm:
I just tried to apply the "TODO", but I'm not sure if it's what was expected.

### DIFF
--- a/ion/src/device/shared/drivers/timing.cpp
+++ b/ion/src/device/shared/drivers/timing.cpp
@@ -10,14 +10,15 @@ namespace Timing {
 
 using namespace Device::Timing;
 
-/* TODO: The delay methods 'msleep' and 'usleep' are currently dependent on the
+/* TODO: The delay method 'usleep' is currently dependent on the
  * optimizations chosen by the compiler. To prevent that and to gain in
  * precision, we could use the controller cycle counter (Systick). */
 
 void msleep(uint32_t ms) {
   // We decrease the AHB clock frequency to save power while sleeping.
+  uint64_t endTime = MillisElapsed + ms;
   Device::Board::setClockFrequency(Device::Board::Frequency::Low);
-  for (volatile uint32_t i=0; i<Config::LoopsPerMillisecond*ms; i++) {
+  while(millis() < endTime) {
     __asm volatile("nop");
   }
   Device::Board::setClockFrequency(Device::Board::standardFrequency());


ion/src/device/shared/drivers/timing.cpp
```diff
@@ -10,14 +10,15 @@ namespace Timing {
...
+/* TODO: The delay method 'usleep' is currently dependent on the
  * optimizations chosen by the compiler. To prevent that and to gain in
  * precision, we could use the controller cycle counter (Systick). */
 
 void msleep(uint32_t ms) {
   // We decrease the AHB clock frequency to save power while sleeping.
   Device::Board::setClockFrequency(Device::Board::Frequency::Low);
-  for (volatile uint32_t i=0; i<Config::LoopsPerMillisecond*ms; i++) {
+  volatile uint64_t endTime = millis() + ms;
+  while(millis() < endTime) {
```
RedGl0w, line 21: Coding style :
```suggestion
  while (millis() < endTime) {
```

RedGl0w (changes requested):
Good idea, however here are other changes to do / questions :
- In the `timing.h` config files, `LoopsPerMillisecond` should be removed
- commits should be squashed
- I added a coding style comment
- Since the sysTick frequency is updated too when the board clock is updated, why not making these change to `usleep` too ?
the systick would need to be changed : we wouldn't want it to be called every 1ms, but every 1us (if it is possible).

Acssiohm:
For the last point , I think it's a good idea too ( but I don't know if it will affect the calculator speed, if so, I don't know if it's worth it )
For the first one, I don't think it's necessary because it's an interesting value that can be used later by others, and if it is not used, the compiler will delete it anyway
Thanks for the coding style comment 
Finally I don't know if you see it , but there is still a bug, that I don't really understand

Acssiohm:
Also, I know I should squash the commits, but , first, I still try to understand the bug, and second I don't know how to do that

RedGl0w:
> For the last point , I think it's a good idea too ( but I don't know if it will affect the calculator speed, if so, I don't know if it's worth it )

The usleep is only used by drivers, so even if there is an important speed difference, it won't be a problem for the end user (if the drivers still works).
> For the first one, I don't think it's necessary because it's an interesting value that can be used later by others, and if it is not used, the compiler will delete it anyway

Even if it is an interessing value, it shouldn't be used anymore, and if we want to find it, the git history still has the value.
> Finally I don't know if you see it , but there is still a bug, that I don't really understand

Yep I was looking about it.

Acssiohm:
>The usleep is only used by drivers, so even if there is an important speed difference, it won't be a problem for the end user (if the drivers still works).
I am not talking about usleep, I'm talking about systick
>the systick would need to be changed : we wouldn't want it to be called every 1ms, but every 1us (if it is possible).
because we will be calling this function a thousand times more often

Acssiohm:
The problem is that it is not us who call the sysTick function, it's an interrupt, when it's configured, you can't change it.

RedGl0w:
> The problem is that it is not us who call the sysTick function, it's an interrupt, when it's configured, you can't change it.

you can set the frequency of systick indeed (and it's configurated : when you go from high speed to low speed, when msleeping, the systick frequency is changed, in order to still tick every 1ms)

Acssiohm:
My bad, if a did understand the datasheet, there is maybe a systick reload value register, that you can change

Acssiohm:
>you can set the frequency of systick indeed (and it's configurated : when you go from high speed to low speed, when msleeping, the systick frequency is changed, in order to still tick every 1ms)

Indeed ... you're right, I didn't see your message before sending mine, and, indeed, the register ( that I was talking about ) is modified in setClockFrequency to compensate ( like you said ) .

Acssiohm:
@RedGl0w I still don't know why this bug appears, but i know what causes it, and how to fix it, but I don't know if it's a good idea to fix it that way.
Anyway, I can do it, and we still can discard the change if it's not preferable to do that

Acssiohm:
You can see that it worked  , but, I had to modify the LD file to make `MillisElapsed` one of the symbols that aren't discarded .
But I wonder why `MillisElapsed` is discarded and why they could use `MillisElapsed` in `millis`but we cannot is `msleep`.
I admit that I don't understand LD files really well.

Acssiohm:
By the way, I know I shouldn't put a .bss in the .data , but I didn't notice at the time, I just wanted to test, if we keep the changes, I'll commit the change

Acssiohm:
I'll just comment the last changes :
- I removed the constant ( just commented it for now ) `LoopsPerMillisecond` like you said
- I made a .bss section for the variable we are using ( because it needs to be in a .bss section, to be initialized correctly )
- And finally I moved `endTime` to the top in case, changing the frequency takes time and removed `volatile` because `endTime` is not volatile ( it is not changed by an outside function, during an interrupt for example )

Acssiohm:
>So you have maybe 2 solutions :
> - you can verify the frequency in the isr_systick() function, to sometimes increment MicrosElapsed ( the variable will need to be renamed ) by 1 , and sometimes by 1000 ( and in millis() you return MicrosElapsed/1000 )
> - or you can just name the variable like sysTicksElapsed and use it in usleep like microseconds, and in msleep like milliseconds , elapsed, by changing the frequency of systicks ...

Wich one do you think is better ? 
( the first require verifying the frequency at every sysTick, the second is a make `millis()` impossible ... )

PR #1869, opened 2021-09-12
[External apps] Allow access to ITCM RAM and reduce slot size

RedGl0w:
All these changes are made in order to enable porting *old* external apps (from zardam) to epsilon 16.


ion/src/device/n0110/shared/config_slot_b.ld
```diff
@@ -1 +1 @@
-EPSILON_OFFSET = 4M;
+EPSILON_OFFSET = 1M;
```
EmilieNumworks, line 1: You're at least overwriting the persisting bytes of slot A.
You may also have issues to execute the external apps and activate the exam mode. I think more changes have to be done if you alter the memory layout given in `ion/src/device/n0110/shared/config_flash.ld`.

RedGl0w: Yep. I knowed when I did that that it wouldn't be necessary. I'm going to try to fix it soon

Issue #2025, opened 2023-02-23
random() implementation is (slightly) flawed


goualard-f:
#### Describe the bug
The `random()` function in the "Calculation" mode uses an algorithm with two flaws:

1/ Contrary to the documentation, it may return the value 1.0 with probability 2^-54
2/ The distribution is slightly non-uniform due to the rounding involved in the `Random::random<>` function (see https://hal.science/hal-02427338).

Possible cure for both problems: in Random::random<>, retain only `p` rightmost bits from `Ion::random()` (with `p` the precision of the type: 24 for `float` and 53 for `double`) and divide by 2^p.

Note: the problem has repercussion for `randint()` as well since `Random::random<>` is used for its implementation. It is even more acute there as `randint()` uses single precision when calling `Random::random<>` (see Issue #2024).

#### Screenshots
Impossible to provided a screenshot since the problem is an elusive one.

#### Expected behavior
`random()` should draw double precision floats with a uniform distribution in [0,1).

#### Environment
 - Epsilon version 20.3.0
 - The platform(s) on which the problem happens: Online simulator and actual device.

PR #2026, opened 2023-02-24
Poincare random improvement

### LINKED ISSUE #2024: randint() implementation is flawed
#### Describe the bug
The `randint()` function in the Calculation module  uses a flawed algorithm that can make it return values outside the requested domain. This can happen in two ways:
1/ When providing bounds that lead to an integer overflow;
2/ When the underlying floating-point RNG returns 1.0f (in the `Random::random<float>` implementation, there are 128 integers out of 2^{32} that can give that result ---approx. 1 in 3e8).

Relying on `Random::random<float>` also entails that `randint()` cannot ensure uniformity: `Ion::random()` generates
integers in [0,2^{32}-1] with a uniform distribution. Once divided by 0xffffffff, some floating-point numbers are created more frequently than others due to rounding (See https://hal.science/hal-02427338). This translates to some integers returned more often than others with `randint()`.

Proposed cure: change the algorithm used entirely (see, e.g., Lemire's nearly divisionless algorithm: https://lemire.me/blog/2019/06/06/nearly-divisionless-random-integer-generation-on-various-systems/).

#### Screenshots
screenshot

#### To Reproduce
1. Go to the 'Calculation' app
2. Type 'randint(0,2^31-1)'
3. See error

This is the way to reproduce the first problem. The second problem, linked to the floating-point RNG cannot easily be reproduced with either the device or the simulator, since it requires calling `randint()` an unknown number of times.

#### Expected behavior

`randint(a,b)` should return an integer in [a,b], not outside of it.

#### Environment
 - Epsilon version (Settings > About > Software version).
20.4.0
 - The platform(s) on which the problem happens: online simulator, actual device, etc...
Online simulator and actual device.

EmilieNumworks:
Attempt to fix #2024 and #2025
The Random::random<T> implementation isn't really satisfying as it samples a small subset of floats (or doubles)...

### DIFF
--- a/omg/include/omg/bit_helper.h
+++ b/omg/include/omg/bit_helper.h
@@ -10,6 +10,7 @@ namespace BitHelper {
 
 constexpr static size_t k_numberOfBitsInByte = 8;
 constexpr static size_t k_numberOfBitsInInt = sizeof(int) * k_numberOfBitsInByte;
+constexpr static size_t k_numberOfBitsInUint32 = sizeof(uint32_t) * k_numberOfBitsInByte;
 
 template <typename T>
 constexpr size_t numberOfBitsInType() {
--- a/poincare/include/poincare/ieee754.h
+++ b/poincare/include/poincare/ieee754.h
@@ -74,16 +74,16 @@ public:
     return exponentBase10;
   }
 
+  constexpr static size_t k_signNbBits = 1;
+  constexpr static size_t k_exponentNbBits = sizeof(T) == sizeof(float) ? 8 : 11;
+  constexpr static size_t k_mantissaNbBits = sizeof(T) == sizeof(float) ? 23 : 52;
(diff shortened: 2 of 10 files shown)


goualard-f (commented):
Concerning the TODO: if you want to be able to draw all the floats in `[0,1)` following a continuous uniform distribution, you have to draw them with a probability proportional to their "basin of attraction*, viz. the interval around each float `x` of all the reals that round to `x`. The earliest unbiased algorithm I know of is presented by [Monahan, 1985](https://www.ams.org/journals/mcom/1985-45-172/S0025-5718-1985-0804945-X/S0025-5718-1985-0804945-X.pdf). There is an earlier algorithm by [Walker, 1974](https://digital-library.theiet.org/content/journals/10.1049/el_19740423) but it draws powers of 2 with the wrong probability.
Anyway, I doubt that such algorithms are really needed for your calculator since $2^{53}$ floats to draw from are usually plenty enough.

PR #2045, opened 2023-04-04
More friendly french translations

Vilerio:


Vilerio:
Here is the explanation of this pull request : in French, the letter "s" means plural, so if the word "produits" is plural, the word "scalaire" must be too, which gives "scalaires".

adriweb:
You're wrong: it's the set of the 3 _individual_ (each singular) products.

artaxxx:
The "s" in "produits" refers both to "scalaire" and "vectoriel", that's why there's an "s"

PR #2051, opened 2023-05-01
Add the ability to draw Gauss curves over histograms

WalrusGumboot:
Hello everyone! This is my first ever pull request on any major open source project, so I'm super excited for it.

For a bit of context: in a stats lesson, my maths teacher said something along the lines of "the only advantage your old TI-84s have over the Numworks that all other students have, is that you're able to draw a Gauss curve over a histogram, to check if the data is normally distributed. Astounded that this wasn't possible in the current version of the software, I decided to try adding it.

I should note that I'm not very acquainted with C++; I mainly write Rust, and in thinking "well, they're both systems languages, so how hard could it be?" I got started. The UI is a bit ugly (I couldn't figure out how to add a switch, so the toggle for the curve is a float which is either 0.0 or 1.0) but the functionality is there.

I'm looking forward to hearing what could be improved and if this could be merged. Thank you for your consideration in any case.


WalrusGumboot (commented):
oh yeah also this fixes a typo in the i18n; BarStartDescrition is now BarStartDescription

EmilieNumworks:
Hi @WalrusGumboot,
Thank you for your proposal! We're discussing the feature internally and come back to you.

alex-kinokon:
@WalrusGumboot Just FYI, this is [not an open-source project](https://github.com/numworks/epsilon/blob/e4c3a98b24c2c407dac3b03332da4fb2c625b1e5/README.md?plain=1#L21).

PR #2073, opened 2023-08-06
Fix Spanish translation in base.es.i18n (issue 2023)

serval2412:
"Decomposição em fatores primos" is Portuguese, not Spanish.


serval2412:
What's the use of Metrics/binary-size? I mean, I've just changed a translation and it fails.
Is there a config pb here?

PiaNumworks:
The translation has already been fix in our private repo, you can see that there is the right text on the beta emulator of version 21.

Metrics/binary-size indicates the size of the binary with your commits and compares it to the previous size. We need to fix it.

PR #2085, opened 2023-09-21
[build] Fix setup: skip check of arm-gcc if --only-simulator

dolmen:
In build/setup.sh, skip check for presence of `arm-none-eabi-gcc` if `--only-simulator` is given as it will not be used.


PiaNumworks:
Thank you for your contribution. We cherry-picked your changes in our private repo to integrate them in the next version of epsilon.

Issue #1987, opened 2022-10-27
Add calorie units under "units -> energy"


aflyingpumpkin:
#### Problem you'd like to fix
Currently there is no way to input calories (or if there is, it is not under units -> energy)

#### Describe the solution you'd like
Add calorie units under "units -> energy"

PR #2123, opened 2023-12-03
Add Calorie Unit to Math Toolbox

Arthur-GYT:
All is in the title ^^

Close https://github.com/numworks/epsilon/issues/1987

I recreated this PR several times because it was my second account that was the author of the PR, which I didn't want.


Arthur-GYT:
@adriweb It's true that there was the rebase or force push, I hadn't thought of that (sorry 😥)

Mysterious-Dev:
@adriweb For your suggestion, I think that the symbol "cal" is better (on Wikipedia, this is marked as "Cal")



https://en.m.wikipedia.org/wiki/Calorie

Mysterious-Dev:
I just realized that I put `ca` and not `Cal` 🤦. Je change ça demain :/

PR #2130, opened 2023-12-26
Move I18n file to specific folders for apps

Arthur-GYT:
All is in the title.

This is to standardize with the `distribution` app and it looks "cleaner".


Arthur-GYT:
Well, I think this PR is ready for review ^^

PR #2143, opened 2024-01-18
[apps/calculation/additional_results] Add missing messages for additional results for complex number

Arthur-GYT:
All is in the title ^^.


PiaNumworks:
Hi @Arthur-GYT,
Thank you for your contribution. The product team has not decided yet whether this change will be included in a future release of Epsilon. We will keep you informed as soon as we have more information.

adri1numworks:
Hello,
Thank you for your suggestion. This is indeed a possibility but it happens very often that the numbers dealt with in this menu are long and this would mean having the message on another line. We would rather avoid this and keep it as simple as possible.
Have a good day !

Issue #2102, opened 2023-10-21
Numpy Arrays Not Vectorizing


HDavidBarth:
#### Describe the bug
Numpy arrays not properly working with scalar multiplication. If you try to multiply a scalar into a numpy array then you get a Type Error: Unsupported types. This only happens if you have the scalar in front of the array, if you have the scalar after the array then it works as expected

#### Screenshots

Expected behavior:
![image](https://github.com/numworks/epsilon/assets/109253006/c0bbd041-2301-4eda-b432-f1364adf7258)

Error:
![screenshot](https://github.com/numworks/epsilon/assets/109253006/9cf63d40-1fc0-4feb-ac9b-9551f38cad99)

Strangely I just realized if you do np.array * scalar then it works, but if you have the scalar in front of the array it does not work properly.

#### To Reproduce
create a numpy arr and multiply by a scalar (scalar in front)

#### Expected behavior
The scalar should multiply by each element in the array and return the scaled array

#### Environment
 - Epsilon version 21.3.0.
 - Online simulator. And handheld

serval2412:
I gave a try with https://github.com/numworks/epsilon/pull/2159 (don't know if it's the right way to fix this or if it'll be accepted but it works for me locally).

PR #2159, opened 2024-02-25
[python] fix multiplication with ndarray when ndarray is rhs

serval2412:
instead of duplicating ndarray management for int/float (more?), just swap lhs and rhs (see bug #2102)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,9 @@ jobs:
       - uses: msys2/setup-msys2@v2
         with:
           update: true
+          install: >-
+            git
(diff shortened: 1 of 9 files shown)


HugoNumworks:
Thanks a lot for giving it a try !

We usually try not to alter micropython's code, especially mixing it with ulab because it makes updating Micropython/Ulab difficult.

I instead found a solution altering Micropython's configuration `MICROPY_PY_REVERSE_SPECIAL_METHODS` to `1`.

It allows a fallback on generic operators from Ulab, and essentially does the same as you did but without altering their code.

PR #2198, opened 2024-05-19
[apps/graph] Add normal to find menu for curve

ao-quantum:
From the graph point explorer -> Find, we already have things like tangent, integral, roots, but no normal.

This PR aims to add a normal line finder, similar to how the tangent works, but it's perpendicular.

I'm pretty new to systems programming, and I don't have a C++ background or foundation, but I did manage to scrape my way through by my knowledge of JS and Rust. I'm open to feedback on anything.

I took a lot of inspiration from how the tangent function works.


HugoNumworks:
Thanks for this contribution !

We're reviewing it with the product team to decide if this could be added to a future release version of Epsilon.
I will keep you updated !

PR #2444, opened 2026-05-01
fix: Binomial exact values

Evangelospro:
It seems that the code is already using the rational definition I proposed. However, the input values were not preserved as exact values. I have created a fix for review and added some test cases. I have tested it using the simulator


poincare/src/expression/arithmetic.cpp
```diff
@@ -321,6 +321,14 @@ bool Arithmetic::ReduceBinomial(Tree* e) {
...
   Tree* n = e->child(0);
   Tree* k = n->nextTree();
+  if (!n->isRational()) {
+    Projection::DeepSystemProject(n);
+    SystematicReduction::DeepReduce(n);
+  }
+  if (!k->isRational()) {
+    Projection::DeepSystemProject(k);
+    SystematicReduction::DeepReduce(k);
+  }
```
HugoNumworks, line 331: These steps should not be necessary. In `ReduceBinomial`, `e` should already be fully projected and every the children have already been reduced.

poincare/src/properties.cpp
```diff
@@ -365,17 +365,24 @@ ComplexProperties Quotient(ComplexProperties p1, ComplexProperties p2) {
...
 
+// Binomial helper: depends on both children (n,k). Implemented in Internal
+// so it can inspect dimensions (lists/matrices) and use Internal helpers.
+ComplexProperties Binomial(const Tree* e) {
+  const Tree* n = e->child(0);
+  const Tree* k = e->child(1);
+  // If any child is a matrix, binomial is not a scalar here.
+  if (Dimension::Get(n).isMatrix() || Dimension::Get(k).isMatrix()) {
+    return ComplexProperties::Unknown();
+  }
```
HugoNumworks, line 378: Dimension check is quite slow, so we try to use it as less as possible, did it solve something in particular ?

HugoNumworks (commented):
Sorry for reviewing this late, and thanks for taking the time to delve into our new math engine.

There are a few things that are not well documented so it is not an easy task when starting from scratch!

If we decide on keeping the feature for a future version, we might cherry-pick the changes in `arithmetic.cpp`, but I'm not sure we need to change the properties for now, unless it fixes something.

Finally, we use a commit naming convention described in https://github.com/numworks/epsilon/blob/master/tools/git/commit-msg and usually add a scope. For example, we would name your first commit something like `feat(poincare): binomial exact values on rational n`

poincare/src/properties.cpp
Evangelospro: I will have a look after the exams again. But I remember there was a test that was failing

